PR #342, opened 2014-04-26
WIP Database refactoring (aka introducing Doctrine)

elkuku:
Some time ago i watched this very interesting [video](https://www.youtube.com/watch?v=C5P59blXuAA) by @HermanPeeren and also read [jooctrine.org](http://jooctrine.org), but life went on and I never got the time to actually play with Doctrine, so I decided to give it a try on my current favorite playground ;)

#### Disclaimer

I have to admit that I

* am no database expert at all
* have never used Doctrine before

so what I am doing here might be just #%&@=!;P

Judge for yourself...

#### This is meant as a "first step":

By
* Start using Doctrine's ORM tool only to setup and update the database tables.
* Keep using (J)DatabaseQuery to query the database.

The immediate benefits would be:
* Get rid of our custom installer code
* Remove the single `mysql.sql` file and move the information to annotations in table classes for every App
* Provide a database update feature

#### Renaming

While we're doing this, it might be a good time to start a discussion on a naming scheme (it has been pointed out somewhere that we are very inconsistent :wink: )

Table names

* `tablename`
* `appname_tablename`
* other...

Table PKs

* `id`
* `tablename_id`
* other...

etc.

Some readings:

* http://leshazlewood.com/software-engineering/sql-style-guide/
* https://launchbylunch.com/posts/2014/Feb/16/sql-naming-conventions/

Both are quite opposite ;)

#### The table prefix (aka `#__`)

<del>Currently the table prefix is required when using the Joomla! database driver and, if not given, defaults to a hard coded `jos_`.</del>

<del>I'm not sure how to handle this with Doctrine, so I decided to hard code it to `_` in the mean time - the `config.json` file has to be adjusted.</del> 
This has been solved.. thanks @HermanPeeren 
but
I *still* wonder if this is something we want to maintain...

#### What's done

Service providers:

* [EntityManager](https://github.com/joomla/jissues/blob/doctrine/src/JTracker/Service/EntityManagerProvider.php)
* [Doctrine console runner](https://github.com/joomla/jissues/blob/doctrine/src/JTracker/Service/DoctrineRunnerProvider.php) (not sure why they did it so complicated creating two custom files...)

New `./bin/jtracker/*` commands:

* `make database` aka `doctrine orm:schema-tool:create`
* `update database` aka `doctrine orm:schema-tool:update`
* `drop database` aka `doctrine orm:schema-tool:drop`
* `test database` aka `doctrine orm:info`

#### What's missing

- [ ] Table keys, foreign key constraints,
- [ ] Initial data import (how this can be done in Doctrine **?** )
- [ ] Remove custom install code and `/etc/mysql.sql` file

I've hacked the `make dbcomments` CLI command a bit to generate the class properties with their annotations. That might be refined.

#### WIP

I'm putting this here for review and discussion.


mbabker:
Personally, as much as I like Doctrine (heck my blog site (custom FW app) was entirely Doctrine versus Joomla Database before I went and made it flat file), I honestly think this starts to stray too far from what Joomla devs & users are comfortable with.  There's already a general reluctance from people to contribute because they have a hard time understanding some of the structure here.

b2z:
Agree with @mbabker

elkuku:
It was a nice experiment, thanks for all the comments, I've learned a lot :)

PR #346, opened 2014-05-02
Move App templates to their proper App directory

elkuku:
This will move the App template files (twig) to a directory of the corresponding App.

This should help in separating the Apps.

**Note** the config changes


mbabker:
We going anywhere with this?

b2z:
I still think that we should keep markup in a separate folder with `theme` parameter:
```
templates/{theme}/*.*
templates/{theme}/{app}/*.*
```
All in one place. Designers do not need to search in apps the markup.

PR #365, opened 2014-05-22
Behat

elkuku:
One of the latest [PHPStorm EAP releases](http://blog.jetbrains.com/phpstorm/2014/05/phpstorm-8-eap-138-184/) contained "initial support" for [Behat](http://behat.org/) which, in conjunction with [Mink](http://mink.behat.org/), can be used to write automated tests. 
I had not heard of Behat before but it immediately caught my attention because of the way tests are written  in natural language -- Like this:

```
  Scenario: Check the top menu "Users"
    Given I am on "/"
    And I follow "Users"
    Then I should see "Users List"
```

This is the whole test. Human readable. No code involved.

Since our test suite is not very extensive currently -- maybe we can explore this a bit.

I also added a method to login a dummy user or a dummy admin for testing purpose. This functionality should only be available in a "local" environment.

Not sure how/if it can be run on Travis. I remember there was the possibility to start some kind of HTTP server?

There is also support for [Selenium2](http://docs.seleniumhq.org/) - just prefix a feature with `@javascript` and fire up the Selenium server.

This is meant as a start :wink:

See also: [Testing Documentation](https://github.com/joomla/jissues/blob/behat/Documentation/Development/Testing.md)


javigomez:
just crossreferencing for system testing. Similar pull but using Codeception instead of Behat/Mink https://github.com/joomla/joomla-cms/pull/6463

Test are not that nice but human readable too: https://github.com/joomla/joomla-cms/pull/6463/files#diff-7392bf22f8e471d7cdd3253559928d2aR27

mbabker:
We need tests.  Of everything.  Can we do something with this?

b2z:
We need a spare time for tests :( Codeception also is nice. Thanks @javigomez for JAB15 session ;) May be sometimes we will have a JTracker Tests Working Group? :+1:

mbabker:
Can't say I've done anything with Behat, but Codeception I've done a little bit with.  The only "issue" with whatever platform gets used for tests around the UI is it's gotta be able to deal with our AJAX'd interface (which I think Codeception does, again don't know enough on Behat to comment there).

PR #551, opened 2014-10-31
Different template path by config file

sebastianluckos:
This will add option to use other template folder - does not use different language/php folders, just for twig.


b2z:
I am closing it in favor of #346

PR #581, opened 2014-12-13
Fix bug in latest filter package. Fixes #579

### LINKED ISSUE #579: Cant select category
Just tried to apply a category to an issue and got the following error

```sql
Cannot add or update a child row: a foreign key constraint fails (`issuesj_tracker`.`ec7hm_issues`, CONSTRAINT `ec7hm_issues_fk_milestone` FOREIGN KEY (`milestone_id`) REFERENCES `ec7hm_tracker_milestones` (`milestone_id`) ON DELETE SET NULL ON UPDATE CASCAD) -- SQL -- UPDATE `ec7hm_issues` SET `issue_number`='5408',`foreign_number`=NULL,`project_id`='1',`milestone_id`='0',`title`='Use the allowed example domain',`description`='<p>This PR change the text to use the correct example domain example.org and also some minor CS issues.</p>',`description_raw`='This PR change the text to use the correct example domain example.org and also some minor CS issues.',`priority`='3',`status`='3',`opened_date`='2014-12-12 07:42:20',`opened_by`='zero-24',`closed_date`='0000-00-00 00:00:00',`closed_by`=NULL,`closed_sha`=NULL,`modified_date`='2014-12-12 23:52:40',`modified_by`='brianteeman',`rel_number`='0',`rel_type`='0',`has_code`='1',`pr_head_user`='',`pr_head_ref`='',`labels`='29',`build`='staging',`easy`='0',`merge_state`='',`gh_merge_status`='',`commits`='' WHERE `id`='6104'
```

wilsonge:
"If the value attribute is not specified, the content will be passed as a value instead." (http://www.w3schools.com/tags/att_option_value.asp - the most reliable source ever).

Michael stated in #579 that the values from the database where

```
["rel_type"] => string(0) ""
["milestone_id"] => string(12) "No Milestone"
```

This fixes that by specifying a default value of 0 for both fields (which isn't a possible relation for the 2 fields). If this value is met then in the model they are set to null (note I don't know whether these checks for 0 are better done in model or table - I was strongly tempted to put them into App\Tracker\Table\IssuesTable::check() function tbh)

Note I have absolutely no virtual box for this thing set up so you MUST test this before merging

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "joomla/date": "~1.1",
         "joomla/di": "~1.1",
         "joomla/event": "~1.1",
-        "joomla/filter": "1.1.2",
+        "joomla/filter": "~1.1",
         "joomla/github": "~1.2",
         "joomla/http": "~1.1",
         "joomla/input": "~1.1",
--- a/src/App/Tracker/Model/IssueModel.php
+++ b/src/App/Tracker/Model/IssueModel.php
@@ -379,6 +379,18 @@ class IssueModel extends AbstractTrackerDatabaseModel
 			throw new \RuntimeException('Missing ID');
 		}
 
+		// If the data rel_type is 0 then we haven't had one given. So make it null in that case.
+		if ($data['rel_type'] === 0)
+		{
+			$data['rel_type'] = null;
+		}
+
+		// If we don't have a milestone given it gets set to 0. So make it null in that case.
+		if ($data['milestone_id'] === 0)
+		{
+			$data['milestone_id'] = null;
+		}
+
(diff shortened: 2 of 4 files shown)


wilsonge:
bump?

elkuku:
I think it's in line with what we are doing currently (checking/filtering in the controller before passing the data to the table class), so I believe we can move this code with the rest when we decide on how to refactor.

@wilsonge honestly this slipped somehow but looking at the submission date that may explain some things... sorry for the delay and thanks for the bump :wink:

wilsonge:
Yeah don't worry - it was that time of year :) Thanks for merging it!!

Issue #586, opened 2015-01-01
Posts deleted at github not deleted on the tracker


brianteeman:
#### Steps to reproduce the issue
Compare https://github.com/joomla/joomla-cms/issues/5148 and http://issues.joomla.org/tracker/joomla-cms/5148

#### Expected result
Spam posts deleted at github are also removed on the issue tracker

#### Actual result
spam posts are not deleted

mbabker:
This one is going to be tricky.  If the API doesn't send some sort of event that indicates a post was deleted nor does it return a response with a message or status code indicating that, I'm not too sure what we could possibly do to remove posts other than manually purge the database.

brianteeman:
Well is it possible to give those of us with permission to delete / edit a
post on GitHub the same abilities on jissues?

mbabker:
I hadn't realized GitHub allowed deleting comments.  Maybe it is possible to update that in the automated handlers as long as there's some sort of API trigger.  As for app permissions, shouldn't be too troublesome to add a delete action.

brianteeman:
Thanks. Really appreciate you looking into this

b2z:
Well I am not sure that we can make it - we do not have any comment ids on the tracker so we can not map them to GitHub comments.

mbabker:
We have the comment ID in the activities table.

b2z:
Yeah thought we did not. Then need to code this feature :laughing: Will try

elkuku:
Honestly, I don't think that this is an enhancement.
We should call it a *critical* bug and try to fix at least the part that removes comments from the JTracker&trade; db when they are removed on GitHub. Otherwise we have to delete spam entries manually from the database...

Let's do the "deleting on JTracker" in a separate PR, unless @b2z feels like he could also code this feature so I wouldn't try to stop him, of course :wink:

b2z:
> Let's do the "deleting on JTracker" in a separate PR

Yeap I will break it into two different tasks:

1. Delete on GH -> delete on JTracker
2. Delete on JTracker -> delete on GH

If you will you can help me out to complete them please let me know which task will you code to avoid duplicate work :)

allenzhao:
I can help on this if no volunteers :)


-------------------------------------------------------------
顺颂时祺  |  Best Wishes

赵泽涵  |  Allen Zhao

中国·对外经济贸易大学 /  信息学院2011级 / 电子商务(国际商务方向)
[...]

Issue #427, opened 2014-08-18
Wrong date for changes in activity


b2z:
During the tests I've noticed that changes has wrong date. See screen:
![111](https://cloud.githubusercontent.com/assets/3051676/3956462/7d5b1090-270b-11e4-99d3-a865e4cded4b.png)

This changes were made on 18th of August, but has 15th.

PR #595, opened 2015-01-15
Add deleting of comments that was deleted on GitHub. Ref #586

b2z:
This will fix the issue when comments deleted on GitHub are not deleted on tracker. Ref #586

### DIFF
--- a/cli/Application/Command/Get/Project/Comments.php
+++ b/cli/Application/Command/Get/Project/Comments.php
@@ -194,6 +194,12 @@ class Comments extends Project
 		// Initialize our ActivitiesTable instance to insert the new record
 		$table = new ActivitiesTable($db);
 
+		// Comments ids for computing the difference
+		$commentsIds = array();
+
+		// Comments ids to delete
+		$toDelete = array();
+
 		// Start processing the comments now
 		foreach ($this->items as $issueNumber => $comments)
 		{
@@ -224,6 +230,9 @@ class Comments extends Project
 
 				foreach ($comments as $i => $comment)
 				{
+					// Store the comment id for computing the difference
+					$commentsIds[] = $comment->id;
+
 					$check = $db->setQuery(
 						$query
 							->clear()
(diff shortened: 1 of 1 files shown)


elkuku:
Yeah this seems to work :wink:

elkuku:
While testing this, I discovered a "thingy" that causes all issues to be processed when our CLI script runs...
https://github.com/joomla/jissues/commit/04d2943dc0f95b0ecc2876f12272d83f0f844892
This has been modified multiple times to fix one bug after another and currently the last modified date is always overwritten...
`@todo`....

b2z:
Can you look also on this PR please #592

b2z:
> While testing this, I discovered a "thingy" that causes all issues to be processed when our CLI script runs...

Hmm... How to avoid it? CLI and not CLI are using the same class. Should we pass some kind of parameter or flag to determine who is acting?

elkuku:
> Should we pass some kind of parameter or flag to determine who is acting?

I think we should not set the last_modified_date in the store method at all... Do you remember the places where it is used by "non CLI" code?...

b2z:
Exactly not, but think that in issue add, save and edit.

elkuku:
@allenzhao added a commit message stating a fix of a modified_date bug in https://github.com/joomla/jissues/commit/dfbb4a5fd2e7c87943beac10f87b0560030d19d9 when he added a check for a db null date but I can't remember what that bug was :(

elkuku:
Also... I feel like something wrong happens here: https://github.com/joomla/jissues/issues/427

b2z:
It seems that we have many places where modified is set :)

elkuku:
I think we are using the last_modified_date from GitHub for issue comments and the date/time on our server for the activities, so I believe there might be a wrong time zone setting here that causes #427 - maybe...

b2z:
Test :wink:

b2z:
@mbabker how often does the cron job run? I've made a comment here more than one hour ago and then deleted it, but still it is [not deleted](http://issues.joomla.org/tracker/jtracker/595#event-61753) on the tracker.

mbabker:
All I'm going to say is it helps when the code is deployed live (that commit was merged on a weekend I was traveling, forgot to sync up when I got back) :blush:

b2z:
glad that it is working after all :tongue:

brianteeman:
Great stuff

Issue #611, opened 2015-02-04
Cron job for CMS not completing


mbabker:
From what I can tell, the cron job for the CMS repo hasn't fully run at least today (I'd need to work through the cron log some more).  It's working fine for all other hosted repos.

Log for the last run of the job:
```sh
[2015-02-04 07:10:01] JTracker.INFO: ---- Bulk Start retrieve Project [] []
[2015-02-04 07:10:01] JTracker.INFO: Processing project: Joomla! CMS [] []
[2015-02-04 07:10:02] JTracker.INFO: Start retrieve Labels [] []
[2015-02-04 07:10:02] JTracker.INFO: Processing project: Joomla! CMS [] []
[2015-02-04 07:10:02] JTracker.INFO: Labels: 0 new, 0 updated, 0 deleted. [] []
[2015-02-04 07:10:02] JTracker.INFO: Finished [] []
[2015-02-04 07:10:02] JTracker.INFO: Start retrieving Milestones [] []
[2015-02-04 07:10:02] JTracker.INFO: Processing project: Joomla! CMS [] []
[2015-02-04 07:10:02] JTracker.INFO: Milestones: 0 new, 7 updated, 0 deleted. [] []
[2015-02-04 07:10:02] JTracker.INFO: Finished [] []
[2015-02-04 07:10:02] JTracker.INFO: Start retrieve Issues [] []
[2015-02-04 07:10:02] JTracker.INFO: Processing project: Joomla! CMS [] []
[2015-02-04 07:10:46] JTracker.INFO: Retrieved 5968 items from GitHub. [] []
```

The next job was the weblinks repo, which had a full run:
```sh
[2015-02-04 07:15:01] JTracker.INFO: ---- Bulk Start retrieve Project [] []
[2015-02-04 07:15:01] JTracker.INFO: Processing project: Weblinks [] []
[2015-02-04 07:15:02] JTracker.INFO: Start retrieve Labels [] []
[2015-02-04 07:15:02] JTracker.INFO: Processing project: Weblinks [] []
[2015-02-04 07:15:02] JTracker.INFO: Labels: 0 new, 0 updated, 0 deleted. [] []
[2015-02-04 07:15:02] JTracker.INFO: Finished [] []
[2015-02-04 07:15:02] JTracker.INFO: Start retrieving Milestones [] []
[2015-02-04 07:15:02] JTracker.INFO: Processing project: Weblinks [] []
[...]

Issue #607, opened 2015-02-04
Recently Updated ?


brianteeman:
Select the recently updated filter.

I would expect this to sort the issues by the most recently updated.

Not sure what it is sorting by but an issue with ID 4022 is at the top of the list. It is almost as if all issues created after this one are not included in the results

b2z:
Well I do not know why but issues from 4022 to 920 has a `Modified 2015-02-04`. That is why these are on the top of the list. If you click on page 4 then you can see that there are issues starting from 5966. That is weird, but seems that `modified_date` was altered.

brianteeman:
OK so that explains why it happened

Can it be fixed to make this feature usable

On 4 February 2015 at 12:32, Dmitry Rekun <notifications@github.com> wrote:

> Well I do not know why but issues from 4022 to 920 has a Modified
> 2015-02-04. That is why these are on the top of the list. If you click on
> page 4 then you can see that there are issues starting from 5966. That is
> weird, but seems that modified_date was altered.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/joomla/jissues/issues/607#issuecomment-72846928>.
>



-- 
Brian Teeman
Co-founder Joomla! and OpenSourceMatters Inc.
http://brian.teeman.net/

b2z:
I hope @mbabker can look into it and alter the `modified_date` in the DB. I do not see other ways how to fix it :(

mbabker:
It has to be something that was done in the cron jobs.  The modified dates of 160 items on this tracker correspond with the cron that ran at 2015-02-04 13:00:00 GMT and the cron log itself supports this:

```sh
[2015-02-04 07:00:02] JTracker.INFO: Start retrieve Issues [] []
[2015-02-04 07:00:02] JTracker.INFO: Processing project: J!Tracker [] []
[2015-02-04 07:00:07] JTracker.INFO: Retrieved 607 items from GitHub. [] []
[2015-02-04 07:00:42] JTracker.INFO: 0 added, 161 updated. [] []
```

b2z:
omg today these issues have 02-05-2105. Something really weird happening here ...

b2z:
Fixed

PR #620, opened 2015-02-06
Ensure that modified date is always set from GitHub

elkuku:
I believe that this might solve several issues we have currently including #607, #611, #617 and maybe others...

I started thinking about it here: https://github.com/joomla/jissues/pull/595#issuecomment-70272396

### DIFF
--- a/src/App/Tracker/Controller/Issue/Save.php
+++ b/src/App/Tracker/Controller/Issue/Save.php
@@ -99,7 +99,10 @@ class Save extends AbstractTrackerController
 
 			try
 			{
-				$this->updateGitHub($item->issue_number, $data, $state, $oldState);
+				$gitHubResponse = $this->updateGitHub($item->issue_number, $data, $state, $oldState);
+
+				// Set the modified_date from GitHub (important!)
+				$data['modified_date'] = $gitHubResponse->updated_at;
 			}
 			catch (GithubException $exception)
 			{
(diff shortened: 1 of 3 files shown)


mbabker:
Let's see what happens...

elkuku:
> Let's see what happens...

:stuck_out_tongue_winking_eye:

elkuku:
Ah.. I believe the cron for the CMS might fail on the first run since all issues have to be fetched...
But if you check the table, the modified dates for the successfully processed items should be correct, so the next run should complete... let's see :wink:

mbabker:
I'll give it a couple hours to run through.  There are still just over 1800 records with a modified date of today in the database, so that should drop drastically as things sync up.

mbabker:
We're close.  5874/6000 issues in the CMS repo got processed before @jissues-bot hit the API limit.  And there are now only 78 records with a modified date of today/

mbabker:
We really need to take a deeper look at https://developer.github.com/v3/#conditional-requests at some point.

mbabker:
Last job ran successfully on the CMS repo.  I think we're at the point now that things are running smoothly again.  Probably will need another bot when the CMS repo gets close to 10,000 items.

elkuku:
Since we are always fetching the whole issue list, we already have the last modified date, so I don't see where we could probably use those conditional requests...

Glad that things seems to run normal again - thanks for the updates :wink:

mbabker:
One immediate thought is if we logged somewhere the last time the command ran for a given project, then the issues that we retrieve are only those updated since the last run.

PR #639, opened 2015-04-11
Use models in web hooks

mbabker:
Still a work in progress, this branch will have the web hooks updated to insert data to the database via model classes versus the table classes.


mbabker:
OK, I think this should be ready for review now.

b2z:
@mbabker looks good, but I wonder should we process the milestones in the hooks like we do it for labels?

mbabker:
Probably.  But that'd be a separate PR I think to add that functionality.  This just cleans things up a bit so that the hooks are validating data with the same rules as the UI input.

b2z:
Ok then. Let's merge and watch for it :)

PR #643, opened 2015-04-18
Add support for unknown events in cron log. Fix #610

### LINKED ISSUE #610: Unknown events in cron log
Just dumping this here but from the cron log, it looks like there are some events we should be processing but aren't, including:

* labeled
* unlabeled
* renamed
* unassigned

b2z:
This PR will add support for unknown events in cron log. Here is the screen how these events looks like:

screenshot


b2z:
@mbabker in case you missed it - ready for review :wink:

b2z:
@mbabker just :up:

PR #648, opened 2015-05-01
Database schema changes

mbabker:
- Adds unique index to the `#__issues` table on issue number and project ID
- Adds cascade operations to the `#__issue_category_map` table's foreign key to the `#__issues` table

### DIFF
--- a/etc/mysql.sql
+++ b/etc/mysql.sql
@@ -27,7 +27,7 @@ CREATE TABLE IF NOT EXISTS `#__tracker_projects` (
   `gh_editbot_user` varchar(150) NOT NULL COMMENT 'GitHub editbot username',
   `gh_editbot_pass` varchar(150) NOT NULL COMMENT 'GitHub editbot password',
   `ext_tracker_link` varchar(500) NOT NULL COMMENT 'A tracker link format (e.g. http://tracker.com/issue/%d)',
-	`short_title` varchar(50) NOT NULL COMMENT 'Project short title',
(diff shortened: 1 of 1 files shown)


b2z:
So I suppose I cannot test it, right?

mbabker:
You could do it in a local database.  If you delete a record from the `#__issues` table, the associated rows in the `#__issue_category_map` table should also delete with this patch (right now MySQL causes the delete to fail due to foreign key constraints).

For the unique index, apply it then try adding a record that has the same project ID and issue number as an existing row, it should fail.  But you should be able to add several records with the same issue number and different project IDs.

b2z:
All good :+1:

mbabker:
11 issues failed the unique key index when this change was made on the live database.  Wow...

b2z:
Well that means we have some issues in our code. Do you have an info where does it happen?

mbabker:
Not a clue.

b2z:
Sad... Is it on cron job right? May be we need to add more logging there.

mbabker:
The webhooks are received and processed in real-time.  The cron jobs just make sure we stay in sync for changes that don't have hooks.  I think the insert issue is probably coming from the hooks, I just have nothing to support that yet.  I can probably rotate out all the logs on the server and enable the database logger, but that one writes every recorded query and I think causes us to hit our hard drive space limits in 5-7 days (in addition to the other loggers).

Issue #681, opened 2015-07-19
Labeling new items causes multiple events


mbabker:
So good ol' @wilsonge just submitted a pull request on the CMS repo and added a label to the item before submitting it.  Somehow, with all the processing, this results in two events being sent and a duplicate index key getting hit in the database:

```sh
[2015-07-19 18:20:22] JTracker.ERROR: Error adding GitHub pull request joomla/joomla-cms #7477 to the tracker: Duplicate entry '7477-1' for key 'issue_project_index' -- SQL -- INSERT INTO `#__issues` (`issue_number`,`project_id`,`title`,`description`,`description_raw`,`status`,`opened_date`,`opened_by`,`modified_date`,`modified_by`,`has_code`,`build`) VALUES  ('7477','1','Use the mock application in the OAuth tests','<p>Mocks the application object in a few OAuth classes and additionally mocks the input object in the twitter tests</p>','Mocks the application object in a few OAuth classes and additionally mocks the input object in the twitter tests','1','2015-07-19 23:20:20','wilsonge','2015-07-19 23:20:20','wilsonge','1','staging') [] []
[2015-07-19 18:20:23] JTracker.INFO: Added 1 labels to joomla/joomla-cms #7477 [] []
[2015-07-19 18:20:24] JTracker.INFO: Added GitHub pull request joomla/joomla-cms #7477 (Database ID #8506) to the tracker. [] []
[2015-07-19 18:20:24] JTracker.INFO: GitHub item joomla/joomla-cms #7477 already has the PR-staging label. [] []
[2015-07-19 18:20:25] JTracker.INFO: Updated GitHub pull request joomla/joomla-cms #7477 (Database ID #8506) to the tracker. [] []
```

Actionable items from this:

- [ ] Web hooks should send non-200 responses back on failures, hopefully it makes identifying them in the GitHub UI easier (see http://cl.ly/image/2V033A2E2O3h)
- [ ] Web hooks need to better detect conditions where multiple pings try to insert records for the same data set

wilsonge:
Glad to be resident bug finder expert :D

wilsonge:
See #682 for the first part

PR #682, opened 2015-07-20
Send back error message and status code if there's an error in the pulls Hook

wilsonge:
Part of the fix to #681 

Note if there is an exception now when performing the hook action we no longer close the application at that point but return an error message an appropriate status header

### DIFF
--- a/src/App/Tracker/Controller/Hooks/ReceivePullsHook.php
+++ b/src/App/Tracker/Controller/Hooks/ReceivePullsHook.php
@@ -68,12 +68,10 @@ class ReceivePullsHook extends AbstractHookController
 			$result = $this->insertData();
 		}
 
-		if (!$result)
+		if ($result)
 		{
-			$this->getContainer()->get('app')->close();
+			$this->response->message = 'Hook data processed successfully.';
 		}
-
-		$this->response->message = 'Hook data processed successfully.';
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/App/Tracker/Controller/Hooks/ReceivePullsHook.php
```diff
@@ -61,11 +61,16 @@ protected function prepareResponse()
...
 		}
 		else
 		{
-			$this->insertData();
+			$result = $this->insertData();
+		}
+
+		if (!$result)
+		{
+			$this->getContainer()->get('app')->close();
```
mbabker: The parent controller deals with the close.  IMO this should be a different message, not just a blank return (look at the screen cap I attached earlier).

src/JTracker/Controller/AbstractAjaxController.php
```diff
@@ -79,6 +79,20 @@ public function execute()
...
 	/**
+	 * Allows setting the status header into the application
+	 *
+	 * @param   int  $code  The status code to set into the application
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function setHeader($code = 200)
```
mbabker: Bad method name IMO, implying you can set all the app headers when you've coded only support for the status header.

wilsonge: I agree - not sure why I did that - fixed

src/App/Tracker/Controller/Hooks/ReceivePullsHook.php
wilsonge: The close was something that was already done inside the respective functions. But yah I've now let things continue and set the error message to be the same as the log message

PR #684, opened 2015-07-20
Return a response with appropriate error code on errors. REF: #681

wilsonge:
Adds in the same changes as in #682 for the other two hook controllers

### DIFF
--- a/src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
+++ b/src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
@@ -155,15 +155,13 @@ class ReceiveCommentsHook extends AbstractHookController
 		catch (\Exception $e)
 		{
 			$logMessage = sprintf(
-				'Error loading the database for comment %d:%s',
+				'Error loading the database for issue %d:%s',
 				$this->hookData->issue->number,
 				$e->getMessage()
 			);
 			$this->logger->error($logMessage);
 			$this->setStatusCode($e->getCode());
 			$this->response->error = $logMessage;
-
-			return false;
 		}
 
 		// Store was successful, update status
@@ -343,15 +341,13 @@ class ReceiveCommentsHook extends AbstractHookController
 		catch (\Exception $e)
 		{
 			$logMessage = sprintf(
-				'Error loading the database for comment %d:%s',
+				'Error loading the database for issue %d:%s',
 				$this->hookData->issue->number,
 				$e->getMessage()
 			);
 			$this->logger->error($logMessage);
 			$this->response->error = $logMessage;
 			$this->setStatusCode($e->getCode());
-
-			return false;
 		}
 
 		// Store was successful, update status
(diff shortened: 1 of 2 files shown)


src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
```diff
@@ -318,11 +342,16 @@ protected function updateComment($id)
...
+			$logMessage = sprintf(
+				'Error loading the database for comment %d:%s',
+				$this->hookData->issue->number,
+				$e->getMessage()
 			);
+			$this->logger->error($logMessage);
+			$this->response->error = $logMessage;
+			$this->setStatusCode($e->getCode());
+
+			return false;
```
wilsonge: Previously we logged a success message and returned true here - I've changed it to returning false but would appreciate someone reviewing that - unsure if it's related to the event being thrown

src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
```diff
@@ -143,11 +154,16 @@ protected function insertComment()
...
+			$logMessage = sprintf(
+				'Error loading the database for comment %d:%s',
+				$this->hookData->issue->number,
+				$e->getMessage()
 			);
+			$this->logger->error($logMessage);
+			$this->setStatusCode($e->getCode());
+			$this->response->error = $logMessage;
+
+			return false;
```
wilsonge: Previously we logged a success message and returned true here - I've changed it to returning false but would appreciate someone reviewing that - unsure if it's related to the event being thrown

mbabker: The try/catch was more for the table load than the event.  We don't need to return a false here because the only thing that fails is the plugin event, not a "critical" failure as with other conditions.

src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
mbabker: Same here, only fail is the plugin event.

src/App/Tracker/Controller/Hooks/ReceiveCommentsHook.php
wilsonge: OK But do we want a non-200 error response?

mbabker: I'd say no as this doesn't cause a critical error.  We should send non-200 responses only when there's a critical fail (i.e. we couldn't parse the hook, there was a failure inserting the hook's issue/comment into the database, etc.) but let the hook be seen as successfully completing if the only fail comes from optional event listeners.

elkuku:
I just discovered that, depending on the servers error reporting settings, the application does in fact provide some kind of errm... error response.
E.g. for the initial "ping" event that is sent from GitHub, one might get a 

```
{"data":{},"error":"Undefined property: stdClass::$issue","message":""}
```

Obviously this goes with a `200` status code...

I think we should improve our error reporting. We are relying 100% on valid data received from GitHub.

Can we redo this PR without altering the workflow?

mbabker:
Outdated.  Webhook handlers have been updated a bunch since this PR.

PR #685, opened 2015-07-31
Update Joomla! Logo

jneubauer:
might as well use the correct logo ;)

switched logo image for new, correct logo, fixing the TM vs R issue as well with an updated mark.

### DIFF
--- a/templates/support/markdown.index.twig
+++ b/templates/support/markdown.index.twig
@@ -336,14 +336,14 @@ You can apply span-level formatting to the content of each cell using regular Ma
         <div class="row-fluid">
             <div class="span6">
                 <h3>{{ "Code"|_ }}</h3>
-                <pre>![Joomla! Logo](http://repository.joomla.org/images/logos/Horizontal/PNG/Joomla%20Logo%20Horz%20Color%20FLAT.png)</pre>
+                <pre>![Joomla! Logo](http://www.joomla.org/images/logos/Horizontal/PNG/Joomla%20Logo%20Horz%20Color%20FLAT.png)</pre>
             </div>
             <div class="span6">
                 <h3>{{ "Output"|_ }}</h3>
                 <p>
                     <a target="_blank"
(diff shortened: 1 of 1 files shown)


MAT978:
It seems (I just checked with the link you provided) that it's a wrong use of the trademark indica.
if you use:
**[logo] Joomla!**
then you use the Registered Trademark then you MUST use the ® 
Thanks

mbabker:
All the stuff in Chiara's folder when the revised brand manual uses the ® so I'd go with that (that's where I've gone to get graphics as I've needed them).

MAT978:
and more than "because" of Chiara, because the US Patent & Trademark Office is saying to do so:
see Serial N°77224536 and N°77226768 ont the http://www.uspto.gov/
and even: https://docs.joomla.org/Joomla:Brand_Identity_Elements/Official_Logo
(still underconstruction but OK) ;)

jneubauer:
I'll update it - we seem to have a serious problem when the official files being served up from joomla.org don't even have the right marks lol

b2z:
This should be closed?

MATsxm:
would say yes as legally speaking and regarding the obligations related to TMs, everything is ok at this point.

mbabker:
The `joomla_logo_black.jpg` file is gone so we're serving a dead image on our Markdown page right now.  Is https://www.joomla.org/images/logos/Horizontal/PNG/Joomla%20Logo%20Horz%20Color%20FLAT.png right or not?

MATsxm:
The one above is registered then can't be used without the R indica.
Please see: https://docs.joomla.org/Joomla:Brand_Identity_Elements/Official_Logo
and/or
https://drive.google.com/a/community.joomla.org/folderview?id=0B-WMsfRg_wivSV83SzZzZ2tRM2M&usp=sharing&urp=https%3A%2F%2Fdrive.google.com%2Ffolderview%3Fid%3D0B-WMsfRg_w#grid

PR #689, opened 2015-08-06
Load language files after routing is processed.

elkuku:
Fixes an issue when items are accessed directly by URI.

Root cause is the bad design of our User object that is [badly coupled to the Project](https://github.com/joomla/jissues/blob/master/src/JTracker/Authentication/User.php#L124) (blame @elkuku on this..)

You may verify that the bug exists by sailing to http://issues.joomla.org/tracker/jtests/44
 
**Expected**: Issue 44 on the test project opens.
**Actual**: Issue 44 on the CMS project opens.

After applying the patch this should be fixed (hopefully).

### DIFF
--- a/src/JTracker/Application.php
+++ b/src/JTracker/Application.php
@@ -112,8 +112,7 @@ final class Application extends AbstractWebApplication implements ContainerAware
 
 		$this->setContainer($container);
 
-		$this->loadLanguage()
-			->mark('Application started');
+		$this->mark('Application started');
 
 		// Register the global dispatcher
 		$this->setDispatcher(new Dispatcher);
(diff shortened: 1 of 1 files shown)


mbabker:
Yep, all good.  Kinda tricky to duplicate (have to be logged out and not have an existing session already) but all's well now.

mbabker:
Heh, so this change finally 404'd some old links (when the aliases for the CMS & Tracker projects were different).  Just spent the last hour trying to figure out what Google Console was e-mailing me about elevated errors over and updating links in some old issues to kill it :smile:

b2z:
Hehe

elkuku:
Glad to hear that it "works" and sorry for your lost time :(

PR #692, opened 2015-08-10
Add event processing to cli get issues

elkuku:
This will add event handling for the CLI process that get issues (and PRs).

The comment event is not implemented yet since its functionality is not clear to me. From what I can see it tries to add a missing behavior like an event "onAfterIssueSave" or something that triggers changes made by jissues?

I hope this brings a bit closer to unify the handling in hooks and cron jobs.

### DIFF
--- a/cli/Application/Command/Get/Project.php
+++ b/cli/Application/Command/Get/Project.php
@@ -15,6 +15,11 @@ use Application\Command\Get\Project\Labels;
 use Application\Command\Get\Project\Milestones;
 use Application\Command\TrackerCommandOption;
 
+use Joomla\Event\Dispatcher;
+use Joomla\Event\Event;
+
+use JTracker\Database\AbstractDatabaseTable;
+
 /**
  * Class for retrieving issues from GitHub for selected projects
  *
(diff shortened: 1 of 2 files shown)


zero-24:
>The comment event is not implemented yet since its functionality is not clear to me.

On the comment event we listen on every comment and do some checks. So the comment event is just the `trigger` to run some checks (e.g. auto RTC label)

>From what I can see it tries to add a missing behavior like an event "onAfterIssueSave" or something that triggers changes made by jissues?

Yes. But this event needs also run if we have a change (new comment) via Github ;)

elkuku:
Well what is not clear to me is the way the JoomlacmsCommentsListener is supposed to work. Currently it checks for a change made **on jissues** (the status changed to RTC) when a comment is added (or modified perhaps) **on GitHub** not doing anything with the comment but with the issue (which is expect as a parameter)

In other words:
If you change the status to RTC on jissues you have to add a comment so the RTC label will appear on GitHub?
This feels wrong to me :(

zero-24:
>If you change the status to RTC on jissues you have to add a comment so the RTC label will appear on GitHub?

Yes this is the current way. If you have a better way :+1: 

But it is nothing critical. If I or others add the RTC label they add a comment like `RTC based on tests` or anything like this ;)

mbabker:
Separate PR but the RTC label shouldn't be checked on the comment event
(things like category, priority, and status should be editable without
comments).  But, that one gets into a funny scenario where we check
something not on GitHub to set something on GitHub so it's definitely a
little awkward.  As we get events better sorted this might be something to
move to our save handlers and not even be dependent on a webhook to trigger.

elkuku:
Yes I guess that is what I was tying to say :wink: 

Regarding this PR: Do you think that it would make any sense trying to implement a trigger for the comment event listeners in the CLI script? or any triggers at all??

I also have concerns about the additional requests being made since currently the "label check" results are not cached and will result in a lot of additional requests when batch processing issues in a cron job...

PR #694, opened 2015-08-13
Global events

elkuku:
This moves the event handling methods from the hooks controller to the base controller to make them available to all controllers.

<del>Adds a new event `onTestAfterSubmit`</del> - listeners <del>will follow</del> are here: #695, #696 :wink:

Currently open #596 could be implemented using events.
I believe that also the code from #667 could be done using an appropriate event instead of commenting :wink:

PR #695, opened 2015-08-14
JoomlacmsTestsListener

elkuku:
This will add an `onTestAfterSubmit` event and a `JoomlacmsTestsListener` event listener that will create a status on GitHub whenever a test is submitted (or altered) on jissues.

FYI GitHub statuses are those annoying messages that appear for "external services" like Travis.

In our case this would reflect the test results (from JBS) to the GitHub UI along with the Travis (or other) test results.

Example:

usr-tests

The text that appears is fully customizable for every project and  I would beg you to postpone discussions about the wording after and only if this PR has been merged - thanks :wink: 

Depends on #694, closes #596

### DIFF
--- a/src/App/Tracker/Controller/Hooks/Listeners/AbstractListener.php
+++ b/src/App/Tracker/Controller/Hooks/Listeners/AbstractListener.php
@@ -8,7 +8,11 @@
 
 namespace App\Tracker\Controller\Hooks\Listeners;
 
+use App\Projects\TrackerProject;
+
 use Joomla\Github\Github;
+
+use JTracker\Github\DataType\Commit\Status;
 
 use Monolog\Logger;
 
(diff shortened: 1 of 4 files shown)


b2z:
#694 is merged now and we can start to test listeners.

And now here is the problem:
```
PHP Fatal error:  Call to undefined method App\\Tracker\\Controller\\TestR
esult\\Ajax\\Submit::addEventListener() in /home/admin/web/jissues.j4devs.ru/jissues/src/App/Tracker/Controller/TestResult/Ajax/Submi
t.php on line 54
```

`AbstractAjaxController` extends `AbstractTrackerController` that is not aware about hooks. Funny thing is that our `AbstractHookController extends AbstractAjaxController` :smile: We need to sort out our hierarchy here :wink:

elkuku:
This looks like changes from the `master` branch are missing - updated :wink: 

The hooks controllers are extending the base AJAX controller because of the response being in JSON notation... I also scratched my head about this after not seeing the code for a while...

mbabker:
I changed that a while back so that when you're looking at the events in GitHub's admin console the response in there isn't a full HTML page (digging through a couple hundred lines of that to find error messages is a pain).

elkuku:
Yes that makes sense.
TBH I wasn't even aware that GitHub stores all those responses and corresponding payloads... pretty awesome! - thanks again @github :wink:

elkuku:
A diagram of **some** of our controller classes:
diagram

b2z:
Works good now! The only problem is the link `Details` from Github to Tracker. When I click on it I get an error:
```
404 UnexpectedValueException
This project does not exist.
```
Testing here
https://github.com/b2z/jissues-test/pull/15

mbabker:
The link is hardcoded to `issues.joomla.org` and that project doesn't exist on our live site :wink:

b2z:
Oh my :) Did notice that it leads to i.j.o :) Well I think this PR is ready for merge.

PR #747, opened 2016-02-20
Add support for new file types. Fix #744

### LINKED ISSUE #744: Upload server information
At the moment we ask for relevent server information. Since 3.5 we now have an export feature (
image ) can we add support for just uploading that config into jissues please?

b2z:
Pull Request for Issue #744 .

#### Testing Instructions
Proceed to [my test server](http://jtracker.j4devs.ru/tracker/jissues-test) and try to upload supported file types.

### DIFF
--- a/src/App/Tracker/Controller/Upload/Ajax/Put.php
+++ b/src/App/Tracker/Controller/Upload/Ajax/Put.php
@@ -46,7 +46,7 @@ class Put extends AbstractAjaxController
 
 			$alt = $isImage
 				? 'screen shot ' . date('Y-m-d') . ' at ' . date('H i s')
-				: trim($file->getName() . '.' . $file->getExtension());
+				: trim($file->getName()) . '.' . $file->getExtension();
 
 			$data = array(
 				array(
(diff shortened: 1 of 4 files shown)


src/App/Tracker/Controller/Upload/Ajax/Put.php
```diff
@@ -42,20 +42,33 @@ protected function prepareResponse()
 			$destDir    = $application->getProject()->project_id;
 			$fullPath   = $host . '/' . $application->get('system.upload_dir') . '/' . $destDir . '/' . $destName;
 
+			$isImage = (false === strpos($file->getMimetype(), 'image')) ? false : true;
```
mbabker: Just leave this as `$isImage = (false !== strpos($file->getMimetype(), 'image'))`.  It's already a boolean state at this point.

b2z: Made it as `$isImage = strpos($file->getMimetype(), 'image');`
Should work also ;)

mbabker: Well, not really.  In that case, if 'image' is at the first character position `strpos()` will return 0 and if it isn't in the string it'll return false.  PHP treats the two values as the same, so you need the explicit `!== false` check (and it can't be `!=` because thanks to PHP's weak type system you'd end up with a 0 != false comparison which doesn't evaluate right).  Gotta love PHP sometimes.

b2z: Yeah :( Fixed :+1:

mbabker:
Looks fine.  We'll see what happens I guess :wink:

PR #749, opened 2016-02-24
 	Add support for project specific issue templates - Fix #748

### LINKED ISSUE #748: Support importing issue templates, per-project templates
Since GitHub now supports issue templates in repos (https://github.com/blog/2111-issue-and-pull-request-templates) we should add functionality to import those templates from the repositories instead of requiring one template for all projects.

GitHub's API has an endpoint to retrieve a file's contents (https://developer.github.com/v3/repos/contents/#get-contents) and since the issue template can only live in two locations a script wouldn't be too complex (check repo root for their presence or the presence of a `.github` directory, if the template file isn't in root and `.github` exists check for a template in there, otherwise fall back on our current template).

Since there isn't a way to open pull requests from here, those templates don't need to be imported too.

elkuku:
Pull Request for Issue #748 .

#### Summary of Changes

* Fetch issue templates using `jtracker get source`
* Issue templates will be used on opening new issues if they exist, otherwise the standard template is used.

**NOTE** GitHub seems to have some "rules" if a file exists in multiple locations or with multiple extensions. The `.github` folder is searched first, so are files without extension.
This behavior is honored by the PR..

#### Testing Instructions

Code review :wink:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 # Files and folders that should not be pushed to the repo #
 /etc/config.json
 .vagrant/*
+/projects/*
 
 # Apple Files #
 .DS_Store
--- a/cli/Application/Command/Get/Source.php
+++ b/cli/Application/Command/Get/Source.php
@@ -1,0 +1,138 @@
+<?php
+/**
+ * Part of the Joomla! Tracker application.
+ *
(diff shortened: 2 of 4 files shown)


mbabker:
Merged #750 instead.

PR #751, opened 2016-03-01
GitHub statuses

elkuku:
Pull Request for Issue # .

#### Summary of Changes
Some time ago GitHub introduced the [Combined Status API](https://developer.github.com/changes/2014-03-27-combined-status-api/)([API](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref))
Currently we display only the last status in the item view (and its state in the list view)

* Importing statuses has been added (and tested) on CLI import.
* Importing statuses has been added (and **NOT** tested) on Hooks import.

#### Testing Instructions
Merge, push, wait until PRs are updated or run the `get project` command with the `-f` option on the server.


mbabker:
Looks fine at a quick glance.

PR #755, opened 2016-03-12
Log Exceptions per PSR-3 standard

mbabker:
#### Summary of Changes

PSR-3 defines a standard for how Exceptions should be logged by attaching it to the context array of a log message.  Monolog supports this structure.  So, all places where Exceptions are being logged now attach the Exception to the context array (in the case of Monolog, the full Exception's data (type, message, code, and trace) get attached to log messages which actually makes these a bit more useful now.

Also changed a couple of places to chain Exceptions where there was a new Exception being thrown.

#### Testing Instructions

Trigger an error, validate Exception data is logged.  Or read the code and make sure I didn't screw this up.

### DIFF
--- a/src/JTracker/Application.php
+++ b/src/JTracker/Application.php
@@ -260,11 +260,10 @@ final class Application extends AbstractWebApplication implements ContainerAware
 			// Log the error
 			$this->getLogger()->critical(
 				sprintf(
-					'Exception of type %1$s thrown with message %2$s',
-					get_class($exception),
-					$exception->getMessage()
+					'Exception of type %1$s thrown',
+					get_class($exception)
 				),
-				['trace' => $exception->getTraceAsString()]
+				['exception' => $exception]
 			);
 
 			$this->setHeader('Status', 500, true);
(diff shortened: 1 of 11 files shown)


src/JTracker/Application.php
```diff
@@ -260,11 +260,10 @@ protected function doExecute()
...
 			$this->getLogger()->critical(
 				sprintf(
-					'Exception of type %1$s thrown with message %2$s',
-					get_class($exception),
-					$exception->getMessage()
+					'Exception of type %1$s thrown',
+					get_class($exception)
 				),
-				['trace' => $exception->getTraceAsString()]
+				['exception' => $exception]
```
b2z: Do we use `trace` somewhere?

mbabker: We don't use it explicitly, it was a way to show the trace for these uncaught Exceptions but that's rendered by Monolog as part of the exception context.

b2z:
Ok I've reviewed the code and do not see any issues. Will merge and lets see. Testing this would be a tough task to do...

mbabker:
For reference, glance at the last couple entries in https://issues.joomla.org/logs/error and that'll show how something was logged without the exception context and with it.

PR #759, opened 2016-03-14
Compiled assets

mbabker:
#### Summary of Changes

This starts reorganizing and redefining how browser assets are managed and deployed.  To try and sum this up:

- Removes the reliance on a full Bower checkout being in place on a server or when users clone; by default the combined (and checked in) assets will be used
- Uses Bower and Grunt for managing the assets for production
    - If you need to make asset changes, you'll have to run `npm install` and `bower install`, only the former is really a new requirement
    - Run `grunt` to update everything; this will recompile the CSS/JS files from the bower sources into a single source, minify them, and copy other static assets to the places we need them
    - Will also take care of minifying core tracker assets, one standardized tool to rule them all
- Layouts with dependencies to media should only load it when the Twig global `templateDebug` is true.  When this is true, the combined assets are not loaded and everything must be loaded individually (this will require you to have setup Bower and Grunt).  With it in the default false state, layouts won't load individual assets in favor of the compiled source.

#### Testing Instructions

All browser assets should continue to load and function

#### TODO Items

- Still need to combine core assets
- ~~Improve compression on the JavaScript file (I had tried this at one point with comments stripped but IIRC I was getting JavaScript errors on that)~~
- The images folder isn't moved into this new media folder; not so sure this NEEDS to though


b2z:
Switched my testing server to this branch and have an issue:
`/media/js/vendor.min.js net::ERR_INCOMPLETE_CHUNKED_ENCODING`

b2z:
One more issue - when I run `grunt` no files are created in `media/js/vendor` and `media/css/vendor`. No errors when tasks are running :(

mbabker:
Did you run `bower install` first?  You should have a `bower_components` directory at the root now if so (I removed the `.bowerrc` config which moved that folder into the `www/vendor` folder).

mbabker:
On the Chrome error, from the sounds of some searches it may relate to how quickly the response is served.  I added support for Apache directives to cache and compress resources at the server level which may help for subsequent requests.

b2z:
Ok I did `bower install`. Noticed some warnings during it:
```
bower deprecated    Package jquery is using the deprecated component.json
bower invalid-meta  jquery is missing "main" entry in bower.json
bower invalid-meta  jquery is missing "ignore" entry in bower.json
```

Also have an error running `grunt`:
```
Running "bower_concat:all" (bower_concat) task
Fatal error: Can’t detect main file for "blueimp-canvas-to-blob" component. You should explicitly define it via bower_concat’s mainFiles option. See Readme for details.
```

b2z:
Regarding `net::ERR_INCOMPLETE_CHUNKED_ENCODING` - I have Nginx running as a proxy, so all static files are served by it. May be the problem is related to Nginx, but I don't know, because there are no errors in logs.

Unfortunately for unknown reason vagrant is not starting virtual machine on my localhost, so I can't test it locally at the moment :(

mbabker:
You have `blueimp-canvas-to-blob` version 2.1.1 installed?  The main file is there in its Bower files.

The other warnings from Bower aren't a big deal.  Some would probably go away when we get around to updating our media components to current versions.

b2z:
> You have blueimp-canvas-to-blob version 2.1.1 installed? The main file is there in its Bower files.
Shouldn't it be installed on `bower install`?

mbabker:
Yes, should be in `bower_components/blueimp-canvas-to-blob` if all the dependencies resolved correctly.  Just trying to make sure the right version got pulled in (blueimp-canvas-to-blob isn't one of our explicit dependencies but pulled in by the other blueimp packages).

b2z:
It's thers, but 3.3.0:
```
"name": "blueimp-canvas-to-blob",
 "version": "3.3.0"
```

mbabker:
Ahh.  OK, I've added that to our bower.json so it gets locked at the right version (this is the same version we've got deployed right now, better to use that than try to figure out if we're compatible with a major version bump).

b2z:
So now I should do `bower uninstall blueimp-canvas-to-blob` and then `bower install blueimp-canvas-to-blob`?

mbabker:
Yes.  It'll prompt about choosing a version when you do the install, make sure you choose the one for 2.1.1.

elkuku:
I have tested this and can confirm the functionality in *production mode* as well as recompiling assets using Grunt. :white_check_mark: 

Would it be possible to add a few lines to our documentation?
One issue I had (as a not-so-noob) was figuring out that `templateDebug` is modified by the `renderer.debug` config setting and not something in `debug`... would be cool to make this `debug.template`

Is this ready for merge or are there `todo`s to be done?
I'd like to merge it right away because of its awesomeness :)

b2z:
If we are sure that `net::ERR_INCOMPLETE_CHUNKED_ENCODING` will not come up on j.org then it is fine to merge.

mbabker:
This should work fine on the .org servers, some sites serve a lot more stuff than the file that's causing issues on your setup.

As for the debug key, it's changed to `debug.template`.

Added a documentation page on the asset management setup.

And there aren't any blocking TODO items here.  Everything else that can/should be done (update versions, clean things up, etc.) can be done in smaller chunks.

elkuku:
Welcome to the next step of evolution :)

PR #761, opened 2016-03-19
App services layer

mbabker:
#### Summary of Changes

This creates an `AppInterface` and base app classes for each of the apps in the overall application.  Compare this to a really simplified version of Symfony bundles.  Long and short it allows each app to load its services into the application structure.  The loading of routes for the router is moved into this layer instead of directly in the application class now.

This layer should be used to help be able to create a more service injection based approach to things versus a service locator approach that's used now.  It also sets up a way to be able to set up the command line environment to be able to use the apps' APIs the same way the web app does.

#### Testing Instructions

Everything still works.

### DIFF
--- a/src/App/GitHub/GitHubApp.php
+++ b/src/App/GitHub/GitHubApp.php
@@ -1,0 +1,45 @@
+<?php
+/**
+ * Part of the Joomla Tracker's GitHub Application
+ *
+ * @copyright  Copyright (C) 2012 - 2014 Open Source Matters, Inc. All rights reserved.
+ * @license    http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License Version 2 or Later
+ */
+
+namespace App\GitHub;
+
+use Joomla\DI\Container;
+use JTracker\AppInterface;
+
+/**
+ * GitHub app
+ *
+ * @since  1.0
+ */
+class GitHubApp implements AppInterface
+{
+	/**
+	 * Loads services for the component into the application's DI Container
+	 *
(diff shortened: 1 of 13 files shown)


src/JTracker/ApplicationTrait.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @copyright  Copyright (C) 2012 - 2014 Open Source Matters, Inc. All rights reserved.
+ * @license    http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License Version 2 or Later
+ */
+
+namespace JTracker;
+
+/**
+ * Trait defining common methods between application classes
+ */
```
b2z: `@since` tag is missing

b2z:
Code seems fine :+1:

src/App/GitHub/GitHubApp.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @throws  \RuntimeException
+	 */
+	public function loadServices(Container $container)
+	{
+		// Register the component routes
+		$maps = json_decode(file_get_contents(__DIR__ . '/routes.json'), true);
+
+		if (!$maps)
+		{
+			throw new \RuntimeException('Invalid router file for the GitHub app.' . $path, 500);
```
elkuku: The `$path` variable isn't defined here, as well as the other instances.

b2z: Oops, how do I missed that :(

mbabker: Fixed

PR #763, opened 2016-03-20
Update the AJAX overlay on the documentation page

mbabker:
#### Summary of Changes

Updates the styling on the AJAX overlay for the documentation page with styles borrowed from the CMS admin.  Mainly the graphic is changed to the Joomla spinner and the loading overlay div is tweaked to make the main container appear opaque during the loading process.

#### Testing Instructions

Make sure documentation still loads, new overlay design is applied.


elkuku:
Looks pretty cool :wink:

PR #767, opened 2016-03-25
Application services

mbabker:
#### Summary of Changes

This converts the top level application objects into services and uses service injection for booting and executing them.

#### Testing Instructions

CLI & Web apps should still work correctly.


cli/tracker.php
```diff
@@ -35,9 +33,83 @@
...
+					'Exception of type %1$s thrown while booting the application',
+					get_class($e)
+				),
+				['exception' => $e]
+			);
+		}
+		catch (\Exception $nestedException)
+		{
+			// Do nothing, we tried our best
+		}
```
elkuku: Not sure if we should try to do "something" here, maybe just echoing the message of the `$nestedException`?

www/index.php
```diff
@@ -23,21 +23,99 @@
...
+				sprintf(
+					'Exception of type %1$s thrown while booting the application',
+					get_class($e)
+				),
+				['exception' => $e]
+			);
+		}
+		catch (\Exception $nestedException)
+		{
+			// Do nothing, we tried our best
```
elkuku: Do "something"?

www/index.php
```diff
@@ -23,21 +23,99 @@
...
+				'Exception of type %1$s thrown while executing the application',
+				get_class($e)
+			),
+			['exception' => $e]
+		);
+	}
+	catch (\Exception $nestedException)
+	{
+		// Do nothing, we tried our best
+	}
```
elkuku: Do "something"?

elkuku:
There seems to be a lot of useful stuff in here. I'm sorry I really didn't got a chance to get my head around this, until now. Would you mind bringing this to a mergeable state again?

cli/tracker.php
mbabker: IMO it's not a big concern here as the nested exception would be an error writing to the log file.  The actual error causing the script to fail is spit out below.

www/index.php
mbabker: Same argument as in the CLI front controller.  Can't say I'm too concerned about an error writing to the log file; the pertinent error message is displayed on screen.

mbabker:
master branch is merged

Issue #781, opened 2016-05-02
Fully upgrade to utf8mb4


mbabker:
We previously updated some of the database schema to support utf8mb4 but it wasn't a full upgrade nor did the underlying database layer support it.  `joomla/database >= 1.4.0` does support it.  Fully upgrade the app & db to use utf8mb4 now.

PR #783, opened 2016-05-04
Full support for MySQL utf8mb4 (Ref #781)

mbabker:
Pull Request for Issue #781

#### Summary of Changes

This PR brings full support for the MySQL utf8mb4 charset to the application.  Pertinent notes:

- utf8mb4 support is hardcoded enabled, GitHub is sending emoji characters over so there's no real way to NOT require this anymore
- Likewise that bumps the MySQL platform requirements to 5.5.3+ for the MySQL server

#### Database Migration

Run `bin/jtracker database migrate` to execute the migration.

#### Testing Instructions

Validate all data is displayed and processed correctly after conversion.


mbabker:
Updated to use the new migrations feature.  // @elkuku

cli/tracker.php
```diff
@@ -17,6 +17,11 @@
 ini_set('display_errors', 1);
 
 define('JPATH_ROOT', realpath(__DIR__ . '/..'));
+define('JPATH_CONFIGURATION', JPATH_ROOT . '/etc');
```
elkuku: This seems to be missing in master so I believe we should merge the branch ASAP :wink:

mbabker: 👍 on that.  I only noticed because I ran the migration on my local to make sure I copied everything from the issue text correctly.

PR #792, opened 2016-06-11
Create a database migrations API

mbabker:
#### Summary of Changes

Creates an API to be able to migrate the database when changes are applied to the base schema.

#### Testing Instructions

For now this is mostly review/critique since there aren't any migrations to test with.  The API is explained in the Documentation file that's added to try and make it clear how it all works.

### DIFF
--- a/Documentation/Development/Database-Migrations.md
+++ b/Documentation/Development/Database-Migrations.md
@@ -1,0 +1,28 @@
+## Database Migrations
+
+A simple `Migrations` class is used to track database migrations for the issue tracker application.
+
+### File Structure
+
+All schema migration files are stored in the `etc/migrations` folder and use a file name in the format of `YYYYMMDD###` where:
+
+* `YYYYMMDD` is the date the migration was added (for example, for 11 June 2016 this would be 20160611)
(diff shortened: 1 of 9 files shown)


elkuku:
Review: All looks cute :wink: 

Questions:

1. I suppose we can redo #783 and check how this works?
1. Something completely irrelevant but it immediately caught my attention that you can identify one thousand versions with only three digits... I don't think that version `0` looks nice but it might be technically correct so [the statement](https://github.com/joomla/jissues/pull/792/files#diff-876d98c86dd5a2f7389ea86c7b93cb70R10) isn't essentially wrong... 

Merged based on this observations :wink:

mbabker:
> I suppose we can redo #783 and check how this works?

Yep.  Will do that in a moment.

> I don't think that version 0 looks nice but it might be technically correct so the statement isn't essentially wrong...

Technically 000 - 999 is 1000 numbers.  In practical terms, we definitely should only have a max of 999 since we probably won't start at 000.  Then again, if we really get that many in one day, I might have to stop and ask what we did wrong.

PR #796, opened 2016-06-14
Crowdin #2

### LINKED ISSUE #787: Translations - Use Crowdin?
Joomla has started using Crowdin for its translation projects - https://crowdin.com/projects/Joomla

Granted, this happened only very recently while we've been using Transifex for years, there have been a couple recommendations for moving to Crowdin.

From my own perspective getting things set up for the patch tester and evaluating tools, things I see as advantages are:

- Import existing translations, so we won't lose anything
- Puts our resources on "project property" (everything Joomla translation is in one spot)
- A [maintained PHP API](https://github.com/akeneo/php-crowdin-api) as well as a [CLI client](https://github.com/crowdin/crowdin-cli) (whereas 4 years after I started it I still have the only Transifex PHP library)

Migrating basically means:

- Writing replacements for the `get transifex` and `update transifex` CLI commands to work with Crowdin
- Setting up the project on Crowdin

Thoughts?

elkuku:
Pull Request for Issue #787

This is an **alternative** to #795 with less code added :wink: 

#### Summary of Changes

* Methods have been added to allow file exchange with https://crowdin.com
* The library elkuku/crowdin-api has been added to the composer file.
* Properties for crowdin credentials have been added to /etc/config.dist.json

#### Testing Instructions

**Upload new language templates**

    jtracker update languagefiles --provider=crowdin --create

**Upload already existing translated files to crowdin**

    jtracker update languagefiles --provider=crowdin --translations

This should only run once.
    
**Download translated files**

    jtracker get languagefiles --provider=crowdin

**Hint**: I very personally would prefer this one over #795 :rocket: 

fix #787 close #795


cli/Application/Command/Get/Crowdin.php
```diff
@@ -0,0 +1,165 @@
...
+	 * @param   string  $domain     The domain of the extension
+	 *
+	 * @return  void
+	 *
+	 * @since   1.0
+	 * @throws  \Exception
+	 */
+	private function receiveFiles($extension, $domain)
+	{
+		// @temp - List with known "exceptions" - @todo move
```
mbabker: Honestly I'd move this to a `getCrowdinLangTag()` method somewhere and be done with it.  Single resource to update and it takes care of needing longer lines like `$langTag = array_key_exists($language, $langMap) ? $langMap[$language] : substr($language, 0, 2);` to handle things differently.

elkuku: Honestly you are right - as usual...

elkuku:
Note that methods have been consolidated in a B/C braking way here...

A `provider` parameter is now required to execute the commands. E.g.

    jtracker update languagefiles --provider=transifex

cli/Application/Command/Make/Langtemplates.php
```diff
@@ -73,7 +73,7 @@ public function execute()
 		$this->getApplication()->outputTitle('Make Language templates');
 
 		ExtensionHelper::addDomainPath('Core', JPATH_ROOT . '/src');
-		ExtensionHelper::addDomainPath('CoreJS', JPATH_ROOT . '/www/media');
+		ExtensionHelper::addDomainPath('CoreJS', JPATH_ROOT . '/www/media/js');
 		ExtensionHelper::addDomainPath('Template', JPATH_ROOT . '/cache/twig');
```
mbabker: LOL so I'm using this exact same path on the live server for the live Twig caching.  Is this conflicting in any way when I run the language update commands from there?  (I know, unrelated to this PR, but it kinda caught my eye)

mbabker:
> Note that methods have been consolidated in a B/C braking way now...

Well, glad we never tagged a "stable" version 😆 

I'd say sync this with master to make sure the service stuff didn't screw anything up and this should be good to go.

cli/Application/Command/Make/Langtemplates.php
elkuku: Hmm I'm not sure. The Twig path hasn't been changed by this PR, so everything should work.

The Javascript path on the other hand has changed (along with some "excludes" because it failed to run on my system and also hasn't changed since march (when somebody merged #759 claiming he tested it...) Compare the `POT-Creation-Date` on 
https://github.com/joomla/jissues/blob/master/src/JTracker/g11n/templates/JTracker.pot#L11 
and 
https://github.com/joomla/jissues/blob/master/src/JTracker/g11n/templates/JTracker.js.pot#L11

I admit that this "fix" shouldn't have been included in this PR...

mbabker: I knew it was spitting errors but I didn't know enough about how it was all working to fix it, the generated files are still right, and it wasn't causing it to break on the live server, so I left it alone.

I guess I can change the cache path on the live server to make sure things don't conflict.  It was one of those things where I saw it and thought "wait a second, 2 things using the same cache for 2 purposes?".

elkuku: TBH the path isn't very pretty. Since it is a temporary thing it should go into a `temp` folder but since we don't have that I (mis)used the `cache` folder.
Looking at the code I think the script creates a `JTracker` sub directory to generate the language files, so there really shouldn't be any issues if you use the same base path for "real" caching stuff...

elkuku:
Synced.

mbabker:
Seems like it works (did a test fetch from Transifex with this branch and it pulled files correctly).

elkuku:
> did a test fetch from Transifex

Nice test... :stuck_out_tongue_winking_eye:

mbabker:
I went with what I had available, which made sure all the application relative code worked right.  I hope your package is tested and works 😝

elkuku:
> I hope your package is tested

Well it has 100% unit test coverage... but since the only "real" tests are in the real world, the jissues project would have the honour to report the first issue then :)

elkuku:
Hmm @mbabker so you are using the new squash/merge feature GitHub offers?
I'd like to say that I don't think it's a good thing. One of the features I love about git is the "commit often" policy...

mbabker:
Actually I just hit the button twice and didn't realize it squashed it until too late.

mbabker:
I turned off the squash merge option.  I didn't realize that GitHub was defaulting to this now.  Guess I should RTFI once in a while.

elkuku:
Al good - stuff happens :wink:

PR #798, opened 2016-06-18
Use Framework v2 Renderer & View, Upgrade Whoops

mbabker:
#### Summary of Changes

Converts the view/rendering API to use the Framework v2 renderer and view packages and upgrades the filp/whoops package to the current version.  Converts the renderer configuration logic to a service provider and moved out of the base controller.

#### Testing Instructions

Double check everything still renders as expected.

#### What Doesn't Work Anymore?

The pure PHP renderer doesn't work anymore.  Considering those templates are long abandoned I don't see the issue with that.  Mustache and Twig integrations should still be fine.


elkuku:
I tried to triple check and everything still seems to render as it should.

One small exception: The clean cache command does not work anymore because of a "3pd" silently requiring the `joomla/filesystem` package that has been removed by this PR.
Updating the `elkuku/g11n` package will solve this issue :wink: 

Since it removes > 1k lines of code I can't resist to push that merge button :tada:

elkuku:
> The pure PHP renderer doesn't work anymore.

No issue with that :stuck_out_tongue_winking_eye:

mbabker:
> because of a "3pd" silently requiring the `joomla/filesystem` package that has been removed by this PR

This is why you declare your dependencies 😝

Issue #764, opened 2016-03-21
Optimize Page Loading


mbabker:
We've got some "heavy" pages and a lot of assets getting loaded, even with the compiled production stuff.  Let's try to optimize a bit.

These screenshots may not be totally fair to the everyday user, generally I leave browser caching off, but it's a good way to gauge the affect of needing to download all the things.

Tracker listing page:
![screen shot 2016-03-21 at 12 30 48 pm](https://cloud.githubusercontent.com/assets/368545/13925996/d6d8e1c8-ef60-11e5-94b2-2746a628d5f7.png)

Single issue view (CMS 9504):
![screen shot 2016-03-21 at 12 33 37 pm](https://cloud.githubusercontent.com/assets/368545/13926101/3b134002-ef61-11e5-9f06-72c697c84173.png)
![screen shot 2016-03-21 at 12 34 39 pm](https://cloud.githubusercontent.com/assets/368545/13926123/55c98d7a-ef61-11e5-9d58-2f127be7af55.png)

Projects list page:
![screen shot 2016-03-21 at 12 35 37 pm](https://cloud.githubusercontent.com/assets/368545/13926162/82c3f464-ef61-11e5-8660-675f80f79c41.png)

b2z:
Do you think this is actual as we have HTTP/2 almost the standard de facto?

mbabker:
AFAIK Joomla's hosting servers aren't supporting HTTP/2 right now.  Even if they do there are still ways we could optimize things (could we use SVG's for the flags instead of separate GIF's that cause 18 additional resource loads per page?).  Some of it's going to be in optimizing the media resources and some of it's going to be in optimizing the PHP code base.

mbabker:
This is in better shape for now.

PR #801, opened 2016-06-21
Language flags

elkuku:
Ref #764 

https://github.com/joomla/jissues/issues/764#issuecomment-211984084
> could we use SVG's for the flags instead of separate GIF's that cause 18 additional resource loads per page

Now this opened not one but several cans of worms for me...

The most easy solution (I found) would be going to https://www.flag-sprites.com generate the combined flag image plus the required CSS and go home.

Unfortunately the flags they offer do not contain some languages we currently use. Mainly catalan and british english. We may leave those blank or just choose a "nearby" one but - using an EU image flag for en-GB might get us in deepest trouble (who knows if they really go with the brexit), and I don't want to get killed for using the "wrong flag"...

So I bit the bullet (before it hits me)... and my first stop was here: http://www.phpied.com/command-line-css-spriting which gave me some ideas on how this could be achieved.

### Language codes

Another thing to consider was the language "code" to use.
Currently we are using "Joomla! language codes&trade;". The only reference I found was at [Micro$oft](https://msdn.microsoft.com/en-us/library/ee825488(v=cs.20).aspx).

There are several other [language codes](https://en.wikipedia.org/wiki/Language_code) including

* [ISO language codes](http://www.lingoes.net/en/translator/langcode.htm)
* [ISO 639-X](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
* [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
* [Crowdin language codes](https://crowdin.com/page/api/language-codes) (which seems to be the only one supporting [Klingon](https://en.wikipedia.org/wiki/Klingon_language) (and probably others) - now that's what I call future proof...)

There are also some PHP Libs dealing with language codes:

* https://github.com/matriphe/php-iso-639
* https://github.com/alcohol/iso3166

I decided not to use any of those but instead reinvent the wheel (again) - I love Joomla! you know - and added a list of known languages with "our" codes to the language helper class. See: https://github.com/joomla/jissues/pull/801/files#diff-fd55cb6560b16d21f62ec0d0898c7e17R43

### Flag images

This is the best and most complete flag images collection that I found and that is free of charge (as in 0$) available on the internet:

* http://forum.tsgk.com/viewtopic.php?t=4921

You really wont miss the [Jolly Rogers](https://en.wikipedia.org/wiki/Jolly_Roger) collection including a flag icon for [Calico Jack Rackham](https://en.wikipedia.org/wiki/Calico_Jack): calico jack rackham

### Test instructions

* There is a list of known languages which is copied from here: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
    * Translators: Please review the code [here](https://github.com/joomla/jissues/pull/801/files#diff-fd55cb6560b16d21f62ec0d0898c7e17R43)  and check that the name that will be displayed in the language selector (the `display` property) is **correct**
* There is a new command `jtracker make languageflags`. This command requires [imagemagick](https://www.imagemagick.org) to be installed on your system and the flag images to be stored "somewhere". To test this command you have to download the flag images from [here](http://forum.tsgk.com/viewtopic.php?t=4921) as they are not included in this PR.<br />
This command will do the following:
    1. Create a combined image with flags for the languages currently "installed" in our `JTracker` application using the [montage]() command and store it at the path given by the `--imagefile` option (defaults to `/cache/flags.png`)
    1. Generate a CSS file at the path given by the `--cssfile` option (defaults to `/cache/flags.css`).
* Hopefully you have a local installation so you can check that the language selector displays and works correctly:

Before | After
--- | ---
2016-06-21-161157_1366x768_scrot | 2016-06-21-160858_1366x768_scrot

#### hF


MATsxm:
If it helps,
as reference, I also use 
https://chronoplexsoftware.com/myfamilytree/localisation/languagecodes.htm
(interesting for native names)
and this one:
https://github.com/jlongster/php-product-details/blob/master/localeDetails.class.php

elkuku:
Yes this indeed helps a lot as the information above is only the result of some hours of internet research performed today, so I may be very wrong in several points.

The first link you posted is a list of language codes that are used in the .NET Framework. Those codes do not follow any standard (I guess?), and - is this what we are using in Joomla!?
And then, I am not sure if I should use 中文(中华人民共和国) or 中文 (Zhōngwén) for Chinese, so I would leave the final decision to the native speakers (I can only cope with 3 languages)

The second link also was quite helpful as it states in the first line:

     Holds locale data for all the locales we currently support

So, again, everybody seems to have to bake his/her own locale file for the languages, ISO-XXX codes, currencies, native names and whatnot that are "supported"...

Any further information on this topic would be highly appreciated. :wink:

www/media/css/language.css
```diff
@@ -0,0 +1,24 @@
+.flag {
+	width: 16px !important;
+	height: 10px !important;
```
elkuku: I had to use `!important` here since the bootstrap selector thingy reduces the flag images and without proper CSS skills, this was the only solution I found.

Any CSS guy/gal that could help me out here?

mbabker:
Anyone who wants to help fix our amazing CSS and layouts is welcome to do so.  Otherwise, this works for me, so...

Sandra97:
@elkuku You can try this:
`.ui.dropdown .menu > .item > img .flag {
width:16px;
height:10px;
}`
with a bit of luck, it will work

elkuku:
@Sandra97 I tried but it doesn't seem to work - sorry ;(

Sandra97:
My fault @elkuku 
Could you try without the blank space between img and .flag?
`.ui.dropdown .menu > .item > img.flag`

elkuku:
YES this did the trick - THANKS - and sorry for being so dumb ;(

Sandra97:
You're not!!! 
Happy if it works!

PR #807, opened 2016-06-23
:uk: Typos - easy fix

MATsxm:
:uk: Typos - easy fix
On code review.
Thanks

### DIFF
--- a/cli/Application/Command/Install/Install.php
+++ b/cli/Application/Command/Install/Install.php
@@ -68,7 +68,7 @@ class Install extends TrackerCommand
 			if (!$this->getApplication()->input->get('reinstall'))
 			{
 				$this->out()
-					->out('<fg=black;bg=yellow>' . g11n3t('WARNING: A database has been found !!') . '</fg=black;bg=yellow>')
+					->out('<fg=black;bg=yellow>' . g11n3t('WARNING: A database has been found!') . '</fg=black;bg=yellow>')
 					->out()
 					->out(g11n3t('Do you want to reinstall ?'))
 					->out()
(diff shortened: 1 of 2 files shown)


cli/Application/Command/Install/Install.php
```diff
@@ -68,7 +68,7 @@ public function execute()
 			if (!$this->getApplication()->input->get('reinstall'))
 			{
 				$this->out()
-					->out('<fg=black;bg=yellow>' . g11n3t('WARNING: A database has been found !!') . '</fg=black;bg=yellow>')
+					->out('<fg=black;bg=yellow>' . g11n3t('WARNING: A database has been found!') . '</fg=black;bg=yellow>')
 					->out()
 					->out(g11n3t('Do you want to reinstall ?'))
```
brianteeman: While you are making these changes it would be good to fix this at the same time
There should be no space before the ?

MATsxm: Done thanks - I try to catch them during my t9n process. 
This is the why.

elkuku: @brianteeman they call it [Plenken](https://en.wikipedia.org/wiki/Plenken) and I am very guilty of it... @MATsxm since we are now translating everything you may also want to remove my smileys (`=;)`) and other inconsistencies like `Finished` vs. `Finished.` that are used several times - but only if your time frame permits this - THANKS for your hard work `=;)`

MATsxm: @elkuku no Pleken in fr-FR as we need the space (not for fr-CA) :wink: 
I will continue working on it, step by step, and also "boost" our translators.

PR #808, opened 2016-06-23
Session language

### LINKED ISSUE #806: Loosing the selected language when a page is loaded
Each time the page is reloaded, you loose the current selected language

#### Steps to reproduce the issue
Select any language (example: french) then any menu or action (example in french: `Projet` -> `Montrer`) then the page will be loaded in en-GB adn have to select fr-FR again to display the french version

#### Expected result
Stay in the language selected

#### Note to @mbabker 
Thank you to say 

> NO

elkuku:
Fix #806

**Bug**: The `users language` parameter (if set) overrides the `session language` parameter (if set) causing the current language to be "resetted" to the users parameter language.

This happens if the following criteria are met:

1. You are logged in.
1. You have set a language in you user settings (in jissues)
1. This language is different from the one you are selecting using the language selector.

#### Summary of Changes
The Application class now has a method to get the current language tag.

Now the language code present in the session (if so) is preferred over the user parameter. I **guess** that this should be the right behaviour.

**Note**: While I completely dislike adding even more code to the Application class, I swear that I will move the `loadLanguage()` method to the `LanguageHelper` class "soonish" :tongue: 

#### Testing Instructions

Check that the erroneous behaviour described in #806 is fixed.

@b2z would you mind checking that nothing broke with the user param please?

### DIFF
--- a/src/JTracker/Application.php
+++ b/src/JTracker/Application.php
@@ -89,7 +89,7 @@ final class Application extends AbstractWebApplication implements ContainerAware
 	 * @var    string
 	 * @since  1.0
 	 */
-	private $languageTag;
+	private $languageTag = 'en-GB';
 
 	/**
 	 * Class constructor.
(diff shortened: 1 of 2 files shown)


b2z:
Tested :+1:

PR #811, opened 2016-06-25
Fixes from Sensio audit

mbabker:
#### Summary of Changes

Address items found in [SensioLabsInsight audit](https://insight.sensiolabs.com/projects/1a93532b-bc3d-4b14-bc94-1189581b42a7/analyses/4)

#### Testing Instructions

Review it 😉


src/App/Activity/Model/ProjectactivityModel.php
```diff
@@ -31,8 +31,6 @@ public function getIssueCounts()
 		$query = $db->getQuery(true);
 
 		$periodList  = [1 => 7, 2 => 30, 3 => 90];
-		$periodNames = [1 => 'Weeks', 2 => 'Months', 3 => 'Quarters'];
```
elkuku: Missing translation?

src/App/Activity/Model/TotaluseractivityModel.php
```diff
@@ -58,8 +58,6 @@ public function getTotalActivity()
 		];
 
 		$periodList  = [1 => 7, 2 => 30, 3 => 90];
-		$periodNames = [1 => 'Weeks', 2 => 'Months', 3 => 'Quarters'];
```
elkuku: Missing translation?

mbabker: Wasn't getting used at all actually.  Probably leftover copy/paste when I was pulling code out of the original CMS component.  The translations are all being done in the controllers.

elkuku:
OK then. 

The test failures are due to some strange errors that are popping up now for the language files. I remember we had that same thing some time ago for russian. @b2z: Do you remember how this got fixed?
Seems like an outstanding bug that has never been solved... investigating.

b2z:
Sorry I do not remember what was the case :(

elkuku:
I have deactivated the language testing (for now). Since we have a script that generates the templates, and an online translation service that creates the language files, there should be no problems, hopefully...

PR #813, opened 2016-07-06
Make the pagination "Older" "Never" available for t9n

MATsxm:
Thank you to check if it is done the way it should + I guess I miss something: any .po file to change?
Thanks


mbabker:
Scripts will rebuild the PO files.  Thanks.

PR #823, opened 2016-07-09
Use OAuth client for authentication, build authorization URL via GitHub API object

mbabker:
I didn't look why this code wasn't working before, but it seems to work fine for me locally so let's use it (seems to be the intent previously?).

Also use the GitHub API objects to build the authorization URL instead of our inline code.


elkuku:
After scratching my head for a while I remembered that I did that because of the following scenario:
We have a second "login" button next to the comment box on the issue view.
I remember that the anchor parameter (`#`) got lost using J OAuth so after a successfull login you would have to scroll down to the comment box again...

Does that still happen? is it a problem??

mbabker:
Good question? 😄

elkuku:
Anchor links are still eaten up :(

mbabker:
Darn 😞

mbabker:
So the anchor probably gets lost because the redirect is based on `$application->get('uri.request')` and I don't think the `Uri` class will have that anchor attached unless you requested the page with the anchor to begin with.

elkuku:
Funny, I was looking at the code thinking the same thing....
I *swear* that it was working at some point :P

PR #830, opened 2016-07-12
Refresh User's name and sync with Github profile

gunjanpatel:
#### Summary of Changes
Refresh user's name

#### Testing Instructions
Before it was not possible to refresh the user's name from Github.
After it will be possible to refresh the user's name


elkuku:
Thanks!

PR #831, opened 2016-07-12
[cli] Add new command to get and set user name

gunjanpatel:
#### Summary of Changes
Added CLI app to update user information

#### Testing Instructions
Use this command to get and update all the user's info.
```bash
cli/tracker.php get users
```


elkuku:
Thanks!

PR #832, opened 2016-07-13
Adding option to fetch issue based on status

gunjanpatel:
#### Summary of Changes

It contains enhancement in CLI app.

option-retrive-based-on-status

### DIFF
--- a/cli/Application/Command/Get/Project/Issues.php
+++ b/cli/Application/Command/Get/Project/Issues.php
@@ -53,7 +53,7 @@ class Issues extends Project
 	 *
 	 * @since  1.0
 	 */
-	protected $issuesStatus = array();
+	protected $issueStates = ['open', 'closed'];
 
 	/**
 	 * Github object as a bot account
@@ -118,15 +118,13 @@ class Issues extends Project
 
 		$resp = trim($this->getApplication()->in());
 
-		$this->issuesStatus = array('open', 'closed');
-
 		if (2 == (int) $resp)
 		{
-			$this->issuesStatus = array('open');
+			$this->issueStates = ['open'];
 		}
-		else if (3 == (int) $resp)
+		elseif (3 == (int) $resp)
 		{
-			$this->issuesStatus = array('closed');
+			$this->issueStates = ['closed'];
 		}
 
 		return $this;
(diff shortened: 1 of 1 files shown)


cli/Application/Command/Get/Project/Issues.php
```diff
@@ -47,6 +47,15 @@ class Issues extends Project
 	protected $issues = array();
 
 	/**
+	 * Status of issues.
+	 *
+	 * @var array
+	 *
+	 * @since  1.0
+	 */
+	protected $issuesStatus = array();
```
elkuku: Could you name this `$issueStatuses` or even better `$issueStates`?
Arrays always should be named as plurals :wink: 

We are also using the "new" array syntax `[]`here . We aren't consistent yet, but new code should use the new syntax :stuck_out_tongue_winking_eye: 

Finally you can initialize the array here instead of line #`121`, so I'd propose:

`protected $issueStates = ['open', 'closed'];`

How does that look (and feel)?

elkuku:
Since the script is also supposed to run unattended (as in a cron job), could you add a command line parameter to pass the `--status`?

The error that appears [here](https://travis-ci.org/joomla/jissues/jobs/144355754) just means that you must use `elseif` instead of `else if` :wink:

cli/Application/Command/Get/Project/Issues.php
gunjanpatel: That look and feels very good. I thought to use `[]` before but later think that it might won't work with older php version so type `array()`. :smile:

gunjanpatel:
@elkuku added option `--status` now.

ie. This command is supported now.
```
cli/tracker.php get project --status=closed
```

elkuku:
This looks pretty good :wink: 

@mbabker would have to change the cron call - any comments?

mbabker:
Just remind me what the cron config needs to be so I can update it.  Otherwise no issue here, unless cPanel's cron manager starts throwing a hissy fit later 😆

elkuku:
Thanks @gunjanpatel !

@mbabker 
> Just remind me what the cron config needs to be so I can update it.

I updated the example cron command in the [documentation](https://github.com/joomla/jissues/blob/master/Documentation/Development/CLI-application.md#import-a-project-from-github). Is this similar to what you (we) are using?

I also changed `tracker.php` to `jtracker` - I think it sounds "cooler" :wink:

mbabker:
OK, that's changed.

gunjanpatel:
Thanks.
Indeed. `jtracker` sounds cool.

b2z:
GJ guys

PR #841, opened 2016-07-18
Remove translator credits

elkuku:
Since Crowdin does not support translator credits (see https://github.com/joomla/jissues/issues/787#issuecomment-227022561) and [this here](https://github.com/joomla/jissues/commit/db42c81fb39c08dec161ffe0b4413be390d81638#diff-d1acff4f2c265f2c2def07d000270067R460) looks like nobody has translated those languages, I'd propose to remove translator credits from our system (actually the code is just commented out :wink:)


MATsxm:
@elkuku @mbabker 
Even if I hate badges, do you want me to build a kind of "Thank you" page with translators for each languages?

brianteeman:
Does the same thing exist for everyone who wrote the code, tested the code
and/or fixed the code?

mbabker:
You kind of get that with the contributor charts and the commit activity (at least for stuff hooked up to the issue tracker or the chart for the docs wiki's activity).  Crowdin doesn't have a similar way to extract out individual contributors in the same way.

MATsxm:
and https://github.com/joomla/jissues/graphs/contributors

elkuku:
Well we do not give credits to testers currently.
But I believe that it's better to start somewhere ...

Am 18.07.2016 1:36 nachm. schrieb "Brian Teeman" <notifications@github.com>:

MATsxm:
@elkuku 
you do: https://issues.joomla.org/activity/user/jtracker

elkuku:
There are no "test points" ;)

Am 18.07.2016 2:04 nachm. schrieb "Marc-Antoine Thevenet" <
notifications@github.com>:

MATsxm:
screenshot-issues joomla org 2016-07-18 15-05-48

mbabker:
We support it.  Whether individuals use it is another story.  Same with the patch tester and weblinks repos.

elkuku:
Lol.. I wonder how that person received those points :P

Am 18.07.2016 2:09 nachm. schrieb "Marc-Antoine Thevenet" <
notifications@github.com>:

elkuku:
> do you want me to build a kind of "Thank you" page with translators for each languages?

That could be nice, but I personally would prefer some kind of script so updating any information would only need to run some command...

mbabker:
Not just for us but it'd be really nice if Crowdin offered a way to get contributor info through their API.  Surely we aren't the only folks who like being able to say thanks.

MATsxm:
I will ask/see with them.

PR #844, opened 2016-07-21
Add more info to the profiler output

elkuku:
#### Summary of Changes

This will add some messages about the loaded MVC classes to the profiler output.

This might not be the best place to show this information, but it was easy to hack in :wink: 

#### Testing Instructions

Turn on debugging and have a look a the profiler output.


mbabker:
Probably should go back one day and add some more ways to push data into the debugger, but for a quick and easy hack it works.

PR #846, opened 2016-07-22
Add php-cs-fixer support

mbabker:
#### Summary of Changes

Adds support for the [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) tool as a means to help programmatically review and correct code style issues in PHP code.  This'll cover everything in the `cli` and `src` directories.

#### Testing Instructions

Review the package's documentation to ensure the selected fixers are in line with the project's code style rules and review the changes made to ensure they are consistent with the project's code style rules.

#### What About PHP_CodeSniffer

This is a tool to be used in addition to PHP_CodeSniffer, especially since Joomla's still using PHPCS 1.x which doesn't support auto-fixing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ cache.properties
 build/api/*
 build/logs/*
 current_SHA
+/.php_cs.cache
 
 # Ignore uploads #
 www/uploads/*
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,57 @@
+<?php
+
+return Symfony\CS\Config\Config::create()
+	->setUsingLinter(false)
+	->setUsingCache(true)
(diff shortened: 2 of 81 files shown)


elkuku:
Not sure if I like the comma on the last line of multi line array declarations, but I guess that I don't have to...

I hope the phpcs-2 team gets some result one of these days, I know they're working hard...

mbabker:
The one true benefit to it is when you add elements to an array the last element doesn't end up as part of your diff because a comma was added to the line.  Beyond that it's mostly a matter of preference.

elkuku:
Oh that's true, haven't thought of diffs.
The only benefit I saw so far is when adding new elements to the array by c&p and not having to add the comma... :tongue:

Issue #852, opened 2016-07-31
RTL Support in the UI


mbabker:
Arabic was recently added to Crowdin for translating the issue tracker.  Assuming it or any other RTL language are worked on and we pull them in, we'll need to make sure the UI can properly display pages for these languages.

mbabker:
Arabic's deployed now and the work I've been doing for the template to support RTL display (since the same person also translated the 3.x landing page) is added in now.  All in all things aren't too bad, just some minor quirks (as expected).

MATsxm:
Thanks for the great work!

bosyrih:
thanks for great job. 
I'm Hussein Al-bosyri Arabic JDocs Translations WG,JDocs Working Group.
the work is perfect but one thing if there is a code in the text like source code for any file, the text must be in LTR because All codes in English that's include documents in translation pages, for better visual.
thanks again for your work.  <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/jtracker/852">issues.joomla.org/jtracker/852</a>.</sub>

Sandra97:
Hi Hussein,
If I understand you well, you mean, for example, being able to keep Joomla! 3.6.2 with LTR display in a RTL page, right?

bosyrih:
Hi Sandra
No that's another thing, I mean when there is a source code for any file or any example code, The code must be in LTR,.  because the code in English language always .

brianteeman:
Span lang=en dir=RTL

bosyrih:
if the example code like this
=========================
protected function parse($raw)
{
    $o = new stdClass;
    $o->text = $raw;

    $this->dispatcher->trigger('onContentPrepare', array('com_tracker.markdown', &$o, new JRegistry));

    echo $o->text;
}
==================

in any screen it must be LTR not RTL.

bosyrih:
that's mean left align not right align to the code.

brianteeman:
Sorry I of course meant to type LTR

bosyrih:
yes brianteeman. 
 correct

bosyrih:
there is another issue in https://issues.joomla.org/markdown => table =>Alignment. the second line in table.  colon on the right of the line for value field in LTR language but in RTL language the line is wrong.
| --------- | -----:| this is for RTl but visually it is appear like |:----- |--------- |

bosyrih:
[...]

PR #854, opened 2016-08-01
Add initial support for "RTL" languages

elkuku:
Pull Request for Issue #852  .

#### Summary of Changes
This will add initial support for "RTL" languages.

#### Testing Instructions
Add a new key to one of the existing languages defined in the [LanguageHelper](https://github.com/joomla/jissues/blob/master/src/JTracker/Helper/LanguageHelper.php#L45) class

    'direction' => 'rtl'

Or add a proper RTL language and set the key as shown.

Further improvements depends on input from a native speaker or someone with more knowledge than me on the subject.

### DIFF
--- a/src/JTracker/Helper/LanguageHelper.php
+++ b/src/JTracker/Helper/LanguageHelper.php
@@ -253,4 +253,21 @@ abstract class LanguageHelper
 
 		return $languages;
 	}
+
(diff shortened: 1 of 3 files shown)


mbabker:
It's enough to get things going for now.

PR #909, opened 2016-11-06
Refactor webhooks (Fix #904)

### LINKED ISSUE #904: Optimize Webhooks
Optimize the webhooks to better handle events.
- The `labeled` or `unlabeled` events should only try to update labels on an issue
- The `milestoned` or `unmilestoned` events should only update the item's milestone
- The `assigned` and `unassigned` events should be ignored as unsupported (we aren't tracking assignees)
- For pull requests, the `synchronize` event can update the full record, but we should stop displaying this activity in the timeline (see https://developer.github.com/v3/activity/events/types/#pullrequestevent and GitHub's note about it)
- The `opened`, `closed`, `reopened`, and `edited` events behave as they do now
- We change the code to be a bit more defensive; unknown events shouldn't attempt to be processed by the hooks by default

mbabker:
Pull Request for Issue #904.

#### Summary of Changes

Restructured the issues and pull requests webhook receivers:

- When inserting new records, we will only do so for `opened`, `closed`, and `reopened` events; we have had funky issues with race conditions when processing other event types as the initial one inserting the data
- When updating existing records, there are now different paths based on the action
    - `opened`, `closed`, `reopened`, `edited`, and `synchronize` (for pull requests) continue to behave as before
    - `labeled` and `unlabeled` events will now only update the item's labels versus trying to update all data
    - Other event types will hit a default case and report as unsupported (this includes `milestoned` and `unmilestoned`, there is no logic for updating milestones implemented into our webhook code at all right now so there's no loss here)

The pull request event listener will also now ignore certain actions as these could be problematic; these should only run on full insert/update/edit events.

This also hides `synchronize` events from the issue timeline, the same as GitHub natively does.

This should be a little more defensive against changing webhook events.  We now have to explicitly code in support for new events versus them automagically being processed, plus this will now limit what some activities do (instead of updating all of the data on every event we only process explicit data for each event type).

#### Testing Instructions

All webhook types continue to receive and process data as before.

### DIFF
--- a/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
+++ b/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
@@ -215,7 +215,7 @@ class JoomlacmsPullsListener extends AbstractListener
 		}
 
 		// Only perform these events for open/close/edit/sync events
-		if (in_array($arguments['action'], ['opened', 'closed', 'reopened', 'edited', 'synchronized']))
+		if (in_array($arguments['action'], ['opened', 'closed', 'reopened', 'edited', 'synchronize']))
 		{
 			// Check that pull requests have certain labels
 			$this->checkPullLabels($arguments['hookData'], $arguments['github'], $arguments['logger'], $arguments['project']);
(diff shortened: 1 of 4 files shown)


b2z:
@mbabker sorry that was not able to test it :( Was in plans on this weekend, but you were faster :)

PR #967, opened 2017-05-03
Update dependencies

elkuku:
Composer tells me that:
`Package raveren/kint is abandoned, you should avoid using it. Use kint-php/kint instead.`
This package is only used in the debug thingy.

So I did just that and also update other dependencies.

Please test :wink:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "filp/whoops": "2.*",
         "league/flysystem": "~1.0",
         "monolog/monolog": "1.*",
-        "raveren/kint": "~1.0",
+        "kint-php/kint": "^2.0",
         "adaptive/php-text-difference": "1.*",
         "pulse00/monolog-parser": "^0.0.3"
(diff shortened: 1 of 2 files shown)


zero-24:
> So I did just that and also update other dependencies.

hmm maybe i miss something but i only see the `raveren/kint` => `kint-php/kint` change in your PR did you commit everything?

mbabker:
It's all in the `composer.lock` file.  The `vendor` directory isn't checked in here.

elkuku:
Do me a favor....

Am 03.05.2017 2:35 nachm. schrieb "Michael Babker" <notifications@github.com
>:

It's all in the composer.lock file. The vendor directory isn't checked in
here.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/joomla/jissues/pull/967#issuecomment-299012766>, or mute
the thread

PR #999, opened 2017-12-22
Convert multiline code to single line

photodude:
Pull Request for Issue code was unnecessarily multiline

#### Summary of Changes
Reduce code to a single line

#### Testing Instructions
Code review should be sufficient.


N6REJ:
Maybe change line 379 just to make it cleaner at the same time?
`. '<jdoc:include type="modules" name="position-0" style="none" />' . "\n\t</div>\n";`

photodude:
@N6REJ I think 378-379 has to be done with these "escape characters" because they are used in the `str_replace()` at line 382

Issue #973, opened 2017-06-13
Rebuild frontend asset tooling


mbabker:
Bower is being phased out of the ecosystem.  We need to update our tooling to account for this.

PR #956, opened 2017-04-07
Fix noversion media paths

b2z:
#### Summary of Changes
Fixed path to noversion media files.

#### Testing Instructions
Set template debug to 1. Ensure that media files are loaded correctly.

PR #1023, opened 2018-06-23
Replace Bower in build pipeline

mbabker:
Pull Request for Issue #973

#### Summary of Changes

Bower is deprecated, this moves us off of relying on it in favor of straight up NPM.  For compiling, this uses Webpack with the [Laravel Mix](https://github.com/JeffreyWay/laravel-mix) wrapper (because I can barely figure that out let alone "plain" Webpack but feel free if you want to work that out).

#### Testing Instructions

All the assets load.  `npm run prod` and `npm run dev` (which have no practical differences right now other than minifying files) both work.

#### Environment Info

At this moment I've got Node 8.9.3 and NPM 6.1.0 installed.  For this to work well the requirements need to be fairly recent versions of both (at least new enough to support the lock file, though I also personally use `npm ci` pretty heavily and the `npm audit` feature is pretty useful.

#### Current Configuration

Right now we've got a lot of outdated UI dependencies.  Some of them are updated in order to be able to get things to install correctly, but there are still a lot needing updates.  Add to that we have "hard" dependencies to jQuery 1.x and Bootstrap 2.3.2 by way of our template, the end result is there are some things that aren't really Webpack friendly right now.  So instead of trying to compile a monolithic vendor file for each UI segment, I've combined the resources of each vendor and those get loaded/used as needed instead.  With an environment supporting HTTP/2, the increased number of resources shouldn't be a problem, plus we can serve smaller responses by not throwing the entire UI stack into 2 files loaded on every page.

There are a few dependencies that aren't put into NPM right now, predominantly because they aren't on the NPM registry, or our jQuery UI build being somewhat customized.

More work can be done on optimizing how things are built, the main goal here is to be able to break the Bower dependency and keep things working, while laying infrastructure to keep moving forward.


elkuku:
I have tested this item :white_check_mark: successfully on 2554f7be271f10a0e26c2e81f54f61de4d7c8e5c<br /><br />Cool @mbabker :wink: <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="http://jissues.local/tracker/jtracker/1023">jissues.local/tracker/jtracker/1023</a>.</sub>

elkuku:
So after testing all the JS functionality we have I'm afraid but I cant find any error :wink:

Visuals also align with the live site ;)

A small glitch with the markdown editor; there seems to be some custom CSS missing?

Some observations:

* When running `npm install` the `package.lock.json` file gets changed.
    The tag `"optional": true` is added to some packages.
* When running `npm run prod` all the files under `/www/media/markitup` are changed (including images). At least they appear as modified in `git status`..

System: Some Linux, npm 6.1.0, node 8.11.3

Can we get rid of

* The grunt stuff
* The `bower.json` file :wink:

Question:
Would it make sense if we take the generated files out of the repo?

I'd :heart: that..

<hr />
RTC :tongue:

elkuku:
Oh and the CI errors are not related to this PR, I believe the `line length check` is a great PITA and we should exclude this check in our repo.

elkuku:
Oh... and what if we move all of our custom CSS and JS to `/assets` and completely .gitignore the `/www/media` directory, or maybe even renaming it to `/www/build` and change the refs?

mbabker:
>> A small glitch with the markdown editor; there seems to be some custom CSS missing?
> 
> I don't know if it's custom CSS or a different version or what the case is right now. Still something to play with I guess.

Found it.  We modified the skin from the upstream source files.  So, we just need to make a new skin with our custom CSS to get the display back to what it's supposed to be.

elkuku:
> We modified the skin from the upstream source files

Uh, bad practice...

elkuku:
> Looks to be related to npm/npm#17722.

It is.

> but I can make use of our Jenkins system to build and deploy (just need to get set up to do that).

Yeah I guess that this would add some workload / another thing to look at to your list of duties which must be huge - just leave it as is ;)

elkuku:
Well... seems that the file upload thingy is not doing its thing :(

Not sure if it's something local or what but I guess you might find out what's going on quicker then me (feeling stupid)..

mbabker:
File upload is fixed.

elkuku:
Looks good now.

mbabker:
The compiler transforms things into a full path relative to the domain root, so declarations like `src:url(/media/fonts/vendor/octicons/build/octicons.eot?d038ccbc4a99be24f33a54b482b2422e);` won't work in a subdirectory (that accounts for most of the 404 errors).  Trying to upload something has a similar problem (tries to POST to `http://projects.j4devs.ru/upload/put/` but it should be `http://projects.j4devs.ru/jtracker/upload/put/`).  TBH I'm not sure if there is a way to fix this one without locally editing [this line](https://github.com/joomla/jissues/blob/c7b75820b236ed6df61224d848afbe1c8751f8a1/webpack.mix.js#L7) in the webpack config if you're pushing to a subdirectory.

b2z:
I do not think there is a need to think about subdirectory in scope of this PR as we are not planning to move to subdirectory :)

elkuku:
Does this "fix" #956 ?

mbabker:
It should, yes, since the same files are now always being loaded.

elkuku:
Just a small reminder:
We have some docs anout asset management at https://github.com/joomla/jissues/blob/master/Documentation/Development/Asset-Management.md
so, if **somebody** finds some time on a rainy day... that would be awesome :wink:

PR #1026, opened 2018-06-24
Use PHPCS 2.x ruleset

mbabker:
#### Summary of Changes

- Migrates to PHPCS 2.x ruleset
- Expands line length to 250 characters
- Fixes most issues found by sniffer
- Sets allowed fail count due to upstream bugs checking tab count in closures

#### Testing Instructions

PHPCS command still runs

### DIFF
--- a/src/App/Activity/Model/UseractivityModel.php
+++ b/src/App/Activity/Model/UseractivityModel.php
@@ -101,9 +101,9 @@ class UseractivityModel extends AbstractTrackerDatabaseModel
 		{
 			// This can only filter Tracker and Test activity types
 			$query->where('t.activity_group = ' . $db->quote($type));
-			$query->order('SUM(CASE WHEN t.activity_group = '
-				. $db->quote($type)
-				. ' THEN t.activity_points ELSE 0 END) DESC, SUM(t.activity_points) + (COUNT(c.id) * 5) DESC');
+			$query->order(
+				'SUM(CASE WHEN t.activity_group = ' . $db->quote($type) . ' THEN t.activity_points ELSE 0 END) DESC, SUM(t.activity_points) + (COUNT(c.id) * 5) DESC'
+			);
 		}
 		elseif ($this->state->get('list.activity_type') == 3)
 		{
--- a/src/App/System/Controller/Config/Save.php
+++ b/src/App/System/Controller/Config/Save.php
@@ -28,7 +28,7 @@ class Save extends AbstractTrackerController
 	 */
 	public function execute()
 	{
-		/* @type \JTracker\Application $application */
+		/** @var \JTracker\Application $application */
 		$application = $this->getContainer()->get('app');
 
 		$application->getUser()->authorize('admin');
@@ -59,6 +59,9 @@ class Save extends AbstractTrackerController
 
 		$application->enqueueMessage(g11n3t('The configuration file has been saved.'), 'success')
 			->redirect('/');
+
+		// To silence PHPCS expecting a return
+		return '';
 	}
 
 	/**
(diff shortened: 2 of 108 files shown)


src/App/System/Controller/Config/Save.php
```diff
@@ -59,6 +59,9 @@ public function execute()
 
 		$application->enqueueMessage(g11n3t('The configuration file has been saved.'), 'success')
 			->redirect('/');
+
+		// To silence PHPCS expecting a return
+		return '';
```
elkuku: `Application::redirect()` should return some string :tongue:

src/App/Activity/Model/UseractivityModel.php
```diff
@@ -101,9 +101,9 @@ public function getUserActivity()
 		{
 			// This can only filter Tracker and Test activity types
 			$query->where('t.activity_group = ' . $db->quote($type));
-			$query->order('SUM(CASE WHEN t.activity_group = '
-				. $db->quote($type)
-				. ' THEN t.activity_points ELSE 0 END) DESC, SUM(t.activity_points) + (COUNT(c.id) * 5) DESC');
+			$query->order(
+				'SUM(CASE WHEN t.activity_group = ' . $db->quote($type) . ' THEN t.activity_points ELSE 0 END) DESC, SUM(t.activity_points) + (COUNT(c.id) * 5) DESC'
```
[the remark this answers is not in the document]
mbabker: Multi-line concat is fine.  Multi-line concatenation at every concat point feels awkward and makes the code harder to read (for me at least).

src/App/System/Controller/Config/Save.php
mbabker: Personally, I'd rather deprecate `AbstractWebApplication::redirect()` and force controllers to set a response that triggers a redirect, similar to https://github.com/joomla/framework.joomla.org/blob/package-docs/src/Controller/Documentation/RedirectController.php, but getting that change accepted in the CMS ecosystem has odds somewhere around slim to none (as is the redirect method means you have to put extra logic in your code to either deal with sniffers like PHPCS warning about a missing return or add some kind of extra logic in the method for when tests trigger redirect and don't call `exit(0)` like the real method does).

elkuku:
TBH I am not quite sure about the added `ALLOWED_FAIL_COUNT` here since there are still fixable errors I believe, but we are getting closer to phpcs V3 :wink:

mbabker:
The bulk of the failures are related to https://github.com/joomla/coding-standards/issues/230, the 7 from `TotalActivity.php` are because I still have commented code there.  So, we can't completely fix them all unless we completely ignore the `Joomla.ControlStructures.ControlStructuresBrackets.SpaceBeforeBrace` sniff.

elkuku:
I guess the phpcs2-test branch can also be deleted?

mbabker:
Yep.

src/App/Activity/Model/UseractivityModel.php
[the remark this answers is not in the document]
elkuku: Finally something to disagree - so glad it's nothing important 😉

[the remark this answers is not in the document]
mbabker: Good part is when we get around to using prepared statements for database queries, concatenation like this goes away 😉

src/App/System/Controller/Config/Save.php
elkuku: I like Symfonys way of returning response objects in controller methods, just return a `RedirectResponse` and let the application handle it..

mbabker: That's how I'm doing Framework apps now (well, actually it's more like `$app->setResponse()` and `return true;` because the controller interface defines a boolean return, `¯\_(ツ)_/¯`).  Unfortunately our controllers copied a lot of the glued together logic of the CMS' component helper and view classes setting up an output buffer and views echoing out stuff versus returning it.  And well, fixing this app to do Response objects is no small task given the number of controllers in place now.

PR #1028, opened 2018-08-06
drop-shadow

brianteeman:
Only people with edit privileges would have seen this but i can and it annoys me so i fixed it

### before

chrome_2018-08-06_17-31-17

### after

chrome_2018-08-06_17-31-08


brianteeman:
Thanks for saving my eyes

Issue #1006, opened 2018-02-17
Sample Data 


wilsonge:
#### Steps to reproduce the issue
Install sample sql data

#### Expected result
I can install the issue tracker

#### Actual result
I can't

```
INSERT INTO `#__tracker_projects` (`project_id`, `title`, `alias`, `gh_user`, `gh_project`, `ext_tracker_link`, `short_title`) VALUES
(1, 'Joomla! CMS', 'joomla-cms', 'joomla', 'joomla-cms', 'http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=%d', 'CMS'),
(2, 'J!Tracker', 'jtracker', 'joomla', 'jissues', '', 'J!Tracker'),
(3, 'Joomla! Security', 'joomla-security', '', '', '', 'JSST')
MySQL said: Documentation

#1364 - Field 'gh_editbot_user' doesn't have a default value
```

b2z:
Is it when you try to run get project?

wilsonge:
nope. when i try to run `bin/jtracker install` to setup my db

b2z:
Hmm, I've tried several times to install from scratch and did not hit  this error. As I understand you are not running vagrant setup? If yes then check this one:
https://stackoverflow.com/questions/15438840/mysql-error-1364-field-doesnt-have-a-default-values

alikon:
maybe dued to strict mode different setting
can you check the `STRICT_TRANS_TABLES` with something like
`SHOW VARIABLES LIKE 'sql_mode';`

if you got `STRICT_TRANS_TABLES` 
 than means that if you declare a field not null  like `gh_editbot_user` then it must have a value always

b2z:
@alikon is it something that we should fix in JIssues to avoid such in the feature?

alikon:
with this requirement : 

> MySQL 5.5.3 with InnoDB support (required to support the MySQL utf8mb4 charset)

a quick & easy fix  should be to  set sql_mode to non_strict mode  (the default for 5.5)
` $db->query("SET @@SESSION.sql_mode = '';"); `

PR #1037, opened 2018-10-11
Mysql strict mode 

PhilETaylor:
Pull Request for Issue #1006 and others.

Finally allows install and get of data with Mysql 5.7.23 with default configuration

Might not be EVERYTHING needed - but this was enough to get me installed and a working app running


etc/mysql.sql
```diff
@@ -23,19 +23,19 @@
 -- Table structure for table `#__tracker_projects`
 --
 
-CREATE TABLE IF NOT EXISTS `#__tracker_projects` (
+CREATE TABLE `j_tracker_projects` (
```
mbabker: Fix table name

PR #1053, opened 2019-03-16
Remove Translation Support from the UI

mbabker:
Note: This PR is best reviewed using [this link](https://github.com/joomla/jissues/compare/a4f87db2f7b0f1ebdc1718e98ceba952d0fd6832...remove/language-support) to avoid the commit removing all of the language files

One of the goals of the issue tracker was to help ease folks into the bug tracking process, and part of the way we aimed to support this was by translating the tracker's interface.  While this might have helped with providing translations for metadata around issues or other parts of the website, issue threads themselves have always been in English and are the dominant content on the page.

An informal poll on Glip showed that few users actually use the translated interface (with some indicating they weren't aware it supported translations), including individuals whose first language is not English.  Unfortunately, the way Google Analytics is set up for the site, we are unable to actually determine how many folks may use the tracker in a non-English language.

Therefore, this pull request begins the removal of translation support from the application by performing the following:

- Removes the translation `.po` files
- Removes the CLI infrastructure to support processing these files
- Removes the Crowdin library integration
- Removes the code in both the CLI and web applications handling setting the user preferred language (keeps bootstrapping the g11n library until all of its uses are fully removed to not breaking the app)
- Removes the language info from the debugging system
- Removes the language switcher from the UI
- Removes localization from the datepicker and jQuery validation libraries

Once merged, all text strings should be unwrapped from the g11n library so that it can be fully removed.


franz-wohlkoenig:
i always use english so its okay to remove other langs. Thanks @mbabker

b2z:
Same for me honestly. Will test it on weekend. 

@mbabker can you attach a db to import please? Thank you.

Sandra97:
The only time I use it in French it's to check if the translations are ok. 
I don't think having the interface translated is very useful, as issues are in English anyway.

elkuku:
Even if it makes me sad I agree with this decision.
I did a quick check on web and cli - everything seems fine.


:heavy_check_mark:

elkuku:
;(

Dmitry Rekun <notifications@github.com> schrieb am So., 17. März 2019,
14:31:

PR #1055, opened 2019-04-03
Remove user listing, adjust routes for account management

mbabker:
Once upon a time, it was envisioned there would be fully featured user profiles on the tracker.  That has never come to fruition, and as is that area is basically useless.  Likewise, there isn't really anything worth administrating about user accounts, so there's no point in having admin support to edit accounts.  So, this PR:

- Removes the `/users` page completely
- Changes the route for the `/user/<id>` page to `/account` and removes support for viewing any user
- Changes the route for the `/user/<id>/edit` page to `/account/edit` and removes support for editing any user
- Changes the route for the `/user/<id>/refresh` route to `/account/github-refresh` and removes support for refreshing GitHub data for any user
- Changes the links to user profiles to always link to GitHub


b2z:
You are too quick for me ))

PR #1071, opened 2020-01-06
missing fontsize in advanced dropdowns

N6REJ:
Currently the advance tools dropdowns have no visible text.


mbabker:
This needs to be overwritten in https://github.com/joomla/jissues/blob/master/assets/scss/jtracker.scss and not modified in these files (these are the compiled CSS files for the .org template, only here to support local dev without needing a network connection to the project's CDN).  It also needs to be identified just why on earth the template is zeroing out these font sizes.

mbabker:
Also, the font sizes are zero'ed out by Bootstrap - https://github.com/twbs/bootstrap/blob/v2.3.2/less/button-groups.less - so either there is some incorrect markup structure or some CSS from one of the plugins is not cascading correctly.  Either way, changing the global template CSS here definitely is not the right fix.

N6REJ:
ok.  Sorry for that.  I'll let you handle it, you've a much better grasp on it then I do.

PR #1078, opened 2020-03-10
Use default application to work around error page edge case

wilsonge:
Pull Request for Issue https://github.com/joomla/jissues/issues/1057 .

#### Summary of Changes
Work around for a circular dependency in the error handler when displaying the error page for a project that does not exist. It's not the cleanest thing ever but should do the job.

#### Testing Instructions
Navigate to a project that doesn't exist. Before patch the emergency fallback error handler. After patch a nice happy one


mbabker:
Eh, works until there's something better in place.

PR #1134, opened 2021-08-30
Prevent Fatal Error

PhilETaylor:
@wilsonge a contribution to prevent a fatal error from a mutual friend.


PhilETaylor:
Dont forget the Use statement too

wilsonge:
Yah that part is relatively ok (although thanks to the friend ;) ). I have a (slightly different but same net result) commented out version sitting on the server now. trying to figure out why the exception being called triggers a full dump of everything is the thing i'm trying to debug :)

PR #1197, opened 2023-10-27
Migrate away from bootstrap datepicker

wilsonge:
#### Summary of Changes
Preparation for decoupling from bootstrap 2 to allow an upgrade to BS5. The bootstrap module used doesn't support a higher version than 4

Note the NPM Update has caused slight differences in the way some of the files are compiled. But there's no material difference beyond the jQuery validation plugin has also been upgraded. Release notes here show no major changes https://github.com/jquery-validation/jquery-validation/releases/tag/1.20.0


N6REJ:
nice to see the migration

PR #1198, opened 2023-10-28
Upgrade jQuery + Bootstrap

wilsonge:
#### Summary of Changes
- Start upgrade to BS5 and jQuery 3 for a template visually compatible with the new J4 sites
- Beta version of bootstrap select for BS5 support
- Initial dark mode support 

#### Testing Instructions
N/A


wilsonge:
@zero-24 can you please review the CSP change I've made here - I needed something that would allow the loading of the inline SVGs in bootstrap for example for the select dropdown icons and also the mobile menu icon.

zero-24:
@wilsonge The change to the CSP looks ok to me. :+1:

PR #1311, opened 2026-01-11
Fix typo: 'recieves' should be 'receives'

pashaarshad:
Fixed a spelling error in the database schema comment where 'recieves' was changed to 'receives' in etc/mysql.sql


wilsonge:
Thankyou!

PR #1340, opened 2026-07-23
Add 'media_source' to the file list in JoomlacmsPullsListener

brianteeman:
Since Joomla 6.2 the media-source folder has moved from the build folder to the root

As a result changes in the media-source were not getting the NPM label. 

This should fix that

### DIFF
--- a/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
+++ b/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
@@ -328,7 +328,7 @@ class JoomlacmsPullsListener extends AbstractListener implements SubscriberInter
             'administrator/components/com_media/package.json',
             'administrator/components/com_media/webpack.config.js',
             '^build/media_source',
-            'media_source',
+            '^media_source',
             'build.js',
             'package-lock.json',
             'package.json',
--- a/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
+++ b/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
@@ -328,7 +328,7 @@ class JoomlacmsPullsListener extends AbstractListener implements SubscriberInter
             'administrator/components/com_media/package.json',
             'administrator/components/com_media/webpack.config.js',
             '^build/media_source',
-            'media_source',    
+            'media_source',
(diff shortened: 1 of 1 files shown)


src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
```diff
@@ -328,6 +328,7 @@ class JoomlacmsPullsListener extends AbstractListener implements SubscriberInter
             'administrator/components/com_media/package.json',
             'administrator/components/com_media/webpack.config.js',
             '^build/media_source',
+            'media_source',
```
richard67:
```suggestion
            'media_source',
```
No trailing white space

src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php
```diff
@@ -328,6 +328,7 @@ class JoomlacmsPullsListener extends AbstractListener implements SubscriberInter
...
```
richard67: In the previous line we use the caret symbol `^` to make sure we check if the path begins with `build/media_source`.

Here we don't have the caret, so this pattern would match both `media_source` and `build/media_source`.

The check happens here using preg_match: https://github.com/joomla/jissues/blob/master/src/App/Tracker/Controller/Hooks/Listeners/JoomlacmsPullsListener.php#L566-L581

So the previous line `'^build/media_source',` will be redundant with the change here.

richard67: Maybe we should use the caret also here: `'^media_source',`

brianteeman: i guess either works - we're a bit inconsistent with using the ^ - happy to change to whatever willl get this merged

richard67: I have no merging privilege in this repo here, but I will be happy to ping maintainers.

I'd suggest to use `'^media_source',` for the new entry and keep the old one with the caret, too, so both match only when they are at the beginning of the string.

brianteeman:
This really needs to be fixed before the PBF

brianteeman:
thanks

